PR #5, opened 2017-05-02
Add travis config

katjam:
Does this look better?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,14 @@
+language: node_js
+node_js:
+    - "4"
+
+before_deploy:
+  npm run build
+
+deploy:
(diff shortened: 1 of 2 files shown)


charlesstrange2 (commented):
Ok, so this just builds then maps /dist as a webroot for github pages?

Note, we don't have any vendor stuff and i don't *think* 'npm run build' does an 'npm install'

katjam:
I think it will npm install by default - but we will see when it runs. It will also run the tests before the build and fail if they fail.

katjam:
Just kicked one off... let's see.

charlesstrange2:
ooh! it does travis messages! :-)

PR #6, opened 2017-05-02
0.1/webpack simple

katjam:
This will be an easier starting point for a quick pre-alpha prototype.

Next steps adding CI, sass and tests. I think Sass is already there - just needs a file to read from.


katjam:
Sass appears to be working... now to add tests.

gingerbreadpup:
But does the sass file compile into a css file?

katjam:
@gingerbreadpup It makes pink text in the output of npm run dev - so 'I guess so' but not actually looked at the dist it outputs.

katjam:
hmm looks like the build files are: build.js  build.js.map  logo.png

So unless the css is bundled into the .js then no - it does not (but I am sure we can fix it)

gingerbreadpup (approved):

PR #9, opened 2017-05-03
Add logo

gingerbreadpup:


src/sass/app.scss
```diff
@@ -1,3 +1,28 @@
...
+  left: 0;
+  top: 0;
+  height: 100%;
+  width: 100%;
+  margin: 0 auto;
+}
+
+.content {
+  position: relative;
+  background-color: #f4efef;
```
katjam, line 24: maybe start with vars while we are small?

katjam (approved):
:shipit: Looks good but for future proof (like if we want to start next sprint off by copying some of our assets - then maybe use vars especially for colour before all gets too big.

src/sass/app.scss
gingerbreadpup: yeah of course! Accidentally included this along with the logo. Will be adding vars soon.

PR #13, opened 2017-05-03
0.1/add vars

gingerbreadpup:
Just pushing this up - don't pay too much attention to my css, I'll be refactoring it all but just wanted to get something up as a starting point.

Let me know if the Google fonts aren't working.

Also, the 'You want to add a voucher' heading is there because that's what we decided the heading would be on the paper prototype for that particular screen - and we will be asking the traders to switch from using the paper prototype to the phone version when they get to that stage.


katjam (approved):
Looks ace to me. :shipit:

PR #23, opened 2017-05-08
Changes to voucherBox to make it force a numberpad on mobile things.

charlesstrange2:


src/sass/app.scss
```diff
@@ -21,7 +21,7 @@ h6 {
   margin-top: 1.5rem;
 }
 
-input[type="text"] {
+input[type="text"], input[type="tel"] {
```
katjam, line 24: Assuming this works better than 'number' for us?

katjam (approved):
:shipit:

src/sass/app.scss
charlesstrange2: Number doesn't allow leading zeros it seems

PR #28, opened 2017-05-10
Replace logo jpg with png

gingerbreadpup:
SVGs are already specified in the webpack config along with other image file types, but at the moment it's just not loading SVGs at all and I'm not sure why - so will just push this PNG logo up for now until I can get it to work.


katjam:
Charlie and I discovered last week that webpack only copies assets that are in use by the templates. So you have to import the .svg into the tpl as well... but you might be doing that already.

katjam (approved):
If you have tried with src=./assets/logo.svg already - just :shipit: this.

PR #29, opened 2017-05-11
0.2/styling

gingerbreadpup:
Only look at this on mobile screens - thanks! Still working on ipad/desktop version.


src/components/Tap.vue
```diff
@@ -0,0 +1,125 @@
...
+                </ul>
+            </div>
+
+        </div>
+
+    </main>
+</template>
+
+<script>
+/* Copyright (c) 2017, Alexander Rose Charity (reg. in England and Wales, #00279157) */
```
katjam: Were we taking all of these out?

src/sass/app.scss
```diff
@@ -59,22 +119,44 @@ button {
...
 
 .container {
   position: relative;
   background-color: $arc_bg;
-  border-radius: 10px;
   margin: 0 auto;
   padding: 2em;
-  width: calc(90% - 4em);
   max-width: 900px;
+  overflow:auto;
```
katjam: missing space.

src/sass/app.scss
```diff
@@ -92,25 +174,159 @@ header {
...
+      left: 9px;
+      -webkit-border-radius: 50%;
+      -moz-border-radius: 50%;
+      border-radius: 50%;
+      zoom: 1;
+      filter: alpha(opacity=0);
+      opacity: 0;
+    }
+    &[type=radio]:focus + label::before {
+      -webkit-box-shadow: 0 0 0 4px $arc_focus;
```
katjam: I don't think we need all 3 anymore... but if we did, sass can handle it.

src/sass/app.scss
```diff
@@ -92,25 +174,159 @@ header {
...
+    &[type=radio] + label::after {
+      content: "";
+      border: 10px solid;
+      width: 0;
+      height: 0;
+      position: absolute;
+      top: 9px;
+      left: 9px;
+      -webkit-border-radius: 50%;
+      -moz-border-radius: 50%;
```
katjam: I don't think we need all 3 anymore... but if we did, sass can handle it.

katjam (approved):
Only minor things.

Plus - testing! But I think that is on another card. :)

:rocket:

src/sass/app.scss
```diff
@@ -47,10 +107,10 @@ label {
 button {
   background-color: $arc_rose;
   color: $arc_white;
-  font-size: 0.8em;
+  font-size:1em;
```
charlesstrange2: space?

src/sass/app.scss
charlesstrange2: space?

charlesstrange2 (commented):
suggestion: put your current components in a "pages" directory and keep "components" for "in-page" custom components?

katjam (changes requested):
Second thoughts - I agree with @charlesstrange2 Pages should be pages. Components reserved for in page reusable stuffs. Also I added vue-router as a dependency.

src/components/Tap.vue
gingerbreadpup: I took it out because it's not needed on that screen, or anywhere as far as I'm aware - but if we do need it further down the line I can put it back in.

katjam: But it is still there. I mean the inline copyright

katjam: I meant to say 'Aren't we taking these out?'

charlesstrange2 (approved):
Reckon that's good
I'll deal with the errant copyright statements in a separate commit. 
:shipit:

PR #31, opened 2017-05-14
Moved navbar, added styling for mobiles

gingerbreadpup:
Moved the navbar to the top so we don't have the keyboard issues, Charlie. For the bottom nav to have been effective I also would have wanted the bottom nav to disappear on scroll down, and reappear on scroll up - maybe further down the line we can look into it?

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -4,7 +4,7 @@ var webpack = require('webpack')
 var HtmlWebpackPlugin = require('html-webpack-plugin')
 
 if (process.env.NODE_ENV === 'production') {
-    var publicPath = './';
+    var publicPath = 'https://neontribe.github.io/ARCVMarket/';
 } else {
     var publicPath = '/';
 }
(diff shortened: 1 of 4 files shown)


webpack.config.js
```diff
@@ -4,7 +4,7 @@ var webpack = require('webpack')
 var HtmlWebpackPlugin = require('html-webpack-plugin')
 
 if (process.env.NODE_ENV === 'production') {
-    var publicPath = 'https://neontribe.github.io/ARCVMarket/';
```
charlesstrange2: While you can dev locally with ./ it needs to be the full https to run if pushed to github pages (at the moment - needs fixing); avoid pushing webpack.config.js OR put it back before you do :-)

charlesstrange2 (changes requested):
Code's good, just wtach webpack.config.js changes

webpack.config.js
gingerbreadpup: I don't even recall making this change (maybe it was when we were trying to get stuff to work the other day?) . Will put it back in.

charlesstrange2:
I think it was when we were getting it working for you - It's to do with
the build rather than dev side.

charlesstrange2 (approved):
:shipit:

PR #35, opened 2017-05-16
Add navbar and header component

gingerbreadpup:
Navbar has been moved to the top (yay), as per the new paper prototype. Will add more components tomorrow and will get the right bits showing up on the right pages.


charlesstrange2 (approved):
:shipit:

PR #36, opened 2017-05-17
Add logo component and styling for logo on login page

gingerbreadpup:
Included import stuff on each page so we only import the components we need on each page (eg. logo but no navbar on login screen) and added styling for the login screen. Removed login checkbox following chat with Charlie.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -14,14 +14,16 @@ import Login from './pages/Login.vue'
 
 // Import components
 import Masthead from './components/Masthead.vue'
+import Logo from './components/Logo.vue'
 
 
 // Define routes
 const routes = [
+{ path: '/', component: Account },
 { path: '/account', component: Account },
 { path: '/tap', component: Tap },
 { path: '/payment', component: Payment },
-{ path: '/', component: Login }
+{ path: '/login', component: Login }
 ]
 
 // Create the router instance and pass the 'routes' option
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -4,7 +4,7 @@ var webpack = require('webpack')
 var HtmlWebpackPlugin = require('html-webpack-plugin')
 
 if (process.env.NODE_ENV === 'production') {
-    var publicPath = 'https://neontribe.github.io/ARCVMarket/';
+    var publicPath = './';
 } else {
     var publicPath = '/';
 }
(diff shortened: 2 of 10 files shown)


webpack.config.js
```diff
@@ -4,7 +4,7 @@ var webpack = require('webpack')
 var HtmlWebpackPlugin = require('html-webpack-plugin')
 
 if (process.env.NODE_ENV === 'production') {
-    var publicPath = 'https://neontribe.github.io/ARCVMarket/';
+    var publicPath = './';
```
charlesstrange2, line 7: Whoops!

charlesstrange2 (changes requested):
webpack script!

webpack.config.js
katjam: @gingerbreadpup Why is this causing you issues with local dev? Your publicPath should only be set by it if you are running in production env.

src/main.js
```diff
@@ -14,14 +14,16 @@ import Login from './pages/Login.vue'
 
 // Import components
 import Masthead from './components/Masthead.vue'
+import Logo from './components/Logo.vue'
 
 
 // Define routes
 const routes = [
+{ path: '/', component: Account },
 { path: '/account', component: Account },
```
katjam, line 23: These are supposed to be the same?

katjam (commented):
Other than the rogue url... :shipit: :icecream:

webpack.config.js
gingerbreadpup: @katjam I need to be able to build it so I can test the built version on my phone, because that's the only way I can access my copy of the site on there. Not sure why that is - ask Charlie for a more informative answer. :)

src/main.js
gingerbreadpup: Yeah, I just chose a page which has the navbar on it - but it ought to be the Add a voucher page though, since that's what will display after logging in. I'll change it to that.

webpack.config.js
katjam: Oh - OK. Maybe we can find a work around for you at some point.

gingerbreadpup:
Have just noticed a problem I need to fix - so that only the content of the page (not the header) changes when clicking on a navbar link. Will close this for now.

PR #38, opened 2017-05-18
0.2/121 accordion

gingerbreadpup:
Progress on the payment page. Markup needs to be tightened up but will do that tomorrow. Bed time.

### DIFF
--- a/src/pages/Payment.vue
+++ b/src/pages/Payment.vue
@@ -84,7 +84,7 @@
 
               </div>
 
-              <button>Get a payment</button>
+              <button class="smaller">Get a payment</button>
 
           </div>
 
--- a/src/pages/Payment.vue
+++ b/src/pages/Payment.vue
@@ -1,95 +1,107 @@
 <template>
-  <div>
-      <main class="container" id="payment">
+    <div>
+        <main class="container" id="payment">
 
-          <div class="content fullwidth">
-              <h1>Requested Payments</h1>
+            <div class="content fullwidth">
+                <h1>Requested Payments</h1>
 
-              <div class="accordion">
+                <div class="accordion">
 
-                <!-- Tab header -->
-                <div class="tab thead">
-                  <label for="tab-one">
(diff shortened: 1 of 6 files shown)


src/pages/Payment.vue
```diff
@@ -2,8 +2,87 @@
...
+                        </div>
+                        <div class="row-code">
+                          <div class="code">RVP12345563</div><div class="date">12/1/17</div>
+                        </div>
+                      </label>
+                    </div>
+                  </div>
+                </div>
+
+              </div>
```
katjam: I am guessing these tab rows will be a loop when you have data? @charlesstrange2 Have you defined a json structure you want that data in yet?

src/pages/Tap.vue
```diff
@@ -7,7 +7,6 @@
                 <h1>Type a voucher code</h1>
 
                 <form id="textVoucher" v-on:submit.prevent>
-                    <label for="voucherBox" id="lblVoucherBox">Type a voucher code</label>
```
katjam, line 10: Is it bad form to not have a label at all - even an invisible one for accessibility?

katjam (changes requested):
Looks pretty good to me. Just the label point.

src/pages/Payment.vue
charlesstrange2: Yup, they'll all be vue'd with for-in statements. JSON is cooking. then SQL/Eloquent to make it happen.

katjam: Have you not got the mocks ready for it yet?

charlesstrange2:
nope - been trying to get auth/state working.

src/pages/Tap.vue
gingerbreadpup: It is bad form, yes - meant to put an invisible one back in but I guess I got distracted!

katjam (approved):
:shipit: :sunflower: Good one!

PR #40, opened 2017-05-19
0.2/121 tidy accordion code, fix scrolling issue

gingerbreadpup:


src/pages/Account.vue
```diff
@@ -1,14 +1,102 @@
...
+                                <div>Voucher total</div>
+                                <div>Amount</div>
+                            </div>
+                        </label>
+                    </div>
 
-                <button>Log out</button>
+                    <!-- Tab row -->
+                    <div class="tab row">
+                        <input id="tab-two" type="checkbox" name="tabs">
```
charlesstrange2, line 23: There's going to be a lot of these tabs when we alter it to generate autmagically. when we do that this is likely to be "tab-2" . Is that going to work for you?

charlesstrange2 (commented):
tab names will probably be sequentially serialised. that ok? (see comment)

src/pages/Account.vue
gingerbreadpup: yeah, that's fine - you can use anything you want. :+1:

charlesstrange2 (approved):
:shipit:

PR #42, opened 2017-05-19
0.2/121 voucher list

gingerbreadpup:


gingerbreadpup:
Will add better styling for this later.

src/pages/Payment.vue
```diff
@@ -1,11 +1,58 @@
...
 
+                <p>You can request payment for <strong>3</strong> vouchers.</p>
+
+                <div class="voucher-list">
+
+                    <!-- Tab header -->
+                    <div class="tab thead">
+                        <label for="tab-one">
+                            <div class="row">
+                                <div>Voucher code</div>
```
katjam, line 16: Can you sort this off fixtures before we merge in? Or you need it in now for style tweaking?

gingerbreadpup: Would be good if I can just merge in as it is first, and then I can work with Charlie to hook it up. If that's ok :dancer:

PR #45, opened 2017-05-22
Add user select screen with placeholder content

gingerbreadpup:
I can have a go at hooking this up once the backend is in place for this - but for now here's the template. I chose radio buttons instead of a select dropdown, I think it's easier to use and requires one less click/tap.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,7 +1,7 @@
 <template>
 
     <div id="app">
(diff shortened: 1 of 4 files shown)


katjam (commented):
Looks good. :shipit: 

The endpoint is: 'api/traders' but you do need to be auth'd to see any

PR #48, opened 2017-05-22
0.2/124 tab bar

gingerbreadpup:
Styling tweaks for the tab bar and list pages. Lists should now look good all the way down to tiny mobiles. CSS needs a bit of tidying up but will do that tomorrow evening.


katjam (approved):
Looks ok to me. :shipit:

PR #50, opened 2017-05-24
Check for recVouchers before getting length

katjam:


gingerbreadpup:
yay, thanks.
:shipit:

katjam:
Can you official approve too - before I merge in?

gingerbreadpup (approved):

charlesstrange2 (approved):
:shipit: though i'm still suspicious as to why it doesn't exist when page is rendered.

katjam:
It does exist, just not right at the start. Otherwise you would not have been getting the value at all before I fixed it. That `if` must be running twice.

charlesstrange2:
And yet, in store, it's defined as an empty array... very odd.

katjam:
Empty = without the [0] item, right?

charlesstrange2:
that's a good point! because Vue does odd things when binding a variable!
Mind at rest now.

PR #51, opened 2017-05-24
Merge remote-tracking branch 'remotes/origin/0.2/release' into 0.2/13…

charlesstrange2:
…9-app-page-route-protection-authorisation

Fix login and routing.

Now people can see the app wired up again...
(Next, wire in "user" page)

### DIFF
--- a/fixtures/fixtures.js
+++ b/fixtures/fixtures.js
@@ -51,7 +51,7 @@ var Fixtures = {
 Fixtures.apply = function (mock) {
     //TODO: example regex route;
     //TODO: example route-parsed route;
-    mock.onGet('/traders/1/vouchers').reply(200,this.traderVouchers["1"]);
+    mock.onGet('/traders/1/vouchers').reply(200, this.traderVouchers["1"]);
 
     mock.onPost('/vouchers').reply(function (request) {
         // returns a success regardless!
@@ -62,9 +62,9 @@ Fixtures.apply = function (mock) {
             "invalid": []
         };
         console.log(data);
-        return [200,data];
+        return [200, data];
         });
-    mock.onPost('/login').reply(200,this.oauthUserToken);
+    mock.onPost('/login').reply(200, this.oauthUserToken);
 
     // route to get nicely structured user vouchers.
     mock.onGet('/traders/1/vouchers/history').reply(200, this.traderVoucherHistory["1"]);
--- a/src/main.js
+++ b/src/main.js
@@ -37,7 +37,7 @@ const routes = [
 
 // Create the router instance and pass the 'routes' option
 const router = new VueRouter({
-    routes, // short for routes: routes
+    routes,
     mode: 'history',
     base: '/'
 });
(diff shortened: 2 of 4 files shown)


fixtures/fixtures.js
```diff
@@ -51,7 +51,7 @@ var Fixtures = {
 Fixtures.apply = function (mock) {
     //TODO: example regex route;
     //TODO: example route-parsed route;
-    mock.onGet('/traders/1/vouchers').reply(200,this.traderVouchers["1"]);
+    mock.onGet('/traders/1/vouchers').reply(200, this.traderVouchers["1"]);
```
katjam, line 54: ugh. Why the string?

src/main.js
```diff
@@ -6,35 +6,53 @@ import VueRouter from 'vue-router';
...
+import Scan from './pages/Scan.vue';
+import Upload from './pages/Upload.vue';
+import Payment from './pages/Payment.vue';
+import Login from './pages/Login.vue';
+import User from './pages/User.vue';
 
 // Import components
-import Masthead from './components/Masthead.vue'
-import Logo from './components/Logo.vue'
-import Instructions from './components/Instructions.vue'
```
katjam: Why do pages need `;` and components don't?

src/main.js
```diff
@@ -6,35 +6,53 @@ import VueRouter from 'vue-router';
...
 ];
 
-// Create the router instance and pass the 'routes' option.
+// Create the router instance and pass the 'routes' option
 const router = new VueRouter({
-    routes, // Short for routes: routes.
-    mode: 'history'
+    routes, // short for routes: routes
+    mode: 'history',
+    base: '/'
```
katjam: Is that always true?

src/main.js
```diff
@@ -6,35 +6,53 @@ import VueRouter from 'vue-router';
...
+// Route Guard rules for directing users
+router.beforeEach((to, from, next) => {
+    var auth = Store.netMgr.isAuth();
+    if (!auth && to.meta.auth) {
+        next({
+            path: '/login',
+            query: {redirect: to.fullPath} // TODO : redirect for deep logins
+        });
+    } else if (auth && !to.meta.auth)  {
+        console.log("hello");
```
katjam: hi!

src/pages/Login.vue
```diff
@@ -48,20 +44,9 @@
...
-             *  Watches this.auth to check for changes
-             */
-            auth: function (val) {
-                if (this.auth) {
-                    // TODO : router wiring
-                }
+                remember: true
             }
-        },
+        },,
```
katjam: 2?

src/pages/Login.vue
```diff
@@ -73,7 +58,7 @@
                 };
                 Store.authenticate(userApiCreds, function () {
                     // I don't like this here, but it's the only place it works for now.
-                    this.auth = true;
+                        this.$router.push({path: '/tap'});
```
katjam: Far out!

katjam (changes requested):
Nearly. Pretty good! :balloon:

fixtures/fixtures.js
charlesstrange2: string?

katjam: I mean the ugly ["1"] but that's presumably because it won't accept int keys?

katjam (approved):
:rocket: Now or change /tap to / as we discussed in here.

fixtures/fixtures.js
charlesstrange2: 1 in this cas eis a label key. it happens to be the zero'th element in that array. I was toying with installing a route-parsing plugin to get laravel style route param niceness. It hasn't happened yet, but the naming system stuck.

PR #53, opened 2017-05-24
0.2/add vouchers counter

gingerbreadpup:
So I know that there are a lot of v-ifs in Tap.vue. The idea is that if you haven't added any vouchers, you will only see the 'Add voucher' button. If you have already added some vouchers, it'll show you the 'Get payment' button too, and the 'Add voucher' button will say 'Add another voucher'.

I'm sure it could be written better - I can make a card for it and then maybe Charlie could work on it with me.

As for the rest, I'm in merge conflict hell but I'm going to create this PR anyway so you can see what I've done. :wine_glass:

### DIFF
--- a/fixtures/fixtures.js
+++ b/fixtures/fixtures.js
@@ -51,7 +51,7 @@ var Fixtures = {
 Fixtures.apply = function (mock) {
     //TODO: example regex route;
     //TODO: example route-parsed route;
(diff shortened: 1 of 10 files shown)


charlesstrange2 (approved):
Sheep it! :sheep:-:it:

PR #54, opened 2017-05-24
Fix button width on tablets or smaller screens

gingerbreadpup:


charlesstrange2 (approved):
100%? 100% approved!
:100: :shipit:

PR #56, opened 2017-05-24
add some fixtures for user traders

charlesstrange2:


fixtures/fixtures.js
```diff
@@ -1,6 +1,41 @@
...
+        "1": [
+            {
+                "id": 2,
+                "name": "Kristy Corntop",
+                "pic_url": null,
+                "market_id": 2,
+                "created_at": "2017-05-24 14:19:22",
+                "updated_at": "2017-05-24 14:19:22",
+                "deleted_at": null,
+                "trader_id": 2
```
katjam: Traders don't have trader_id

fixtures/fixtures.js
```diff
@@ -1,6 +1,41 @@
-import Route from 'route-parser';
-
+import Store from '../src/store.js';
 var Fixtures = {
+    userTraders: {
```
katjam, line 3: You mean traders?

katjam (changes requested):

fixtures/fixtures.js
charlesstrange2: traders that belong to a user. ergo userTraders?

fixtures/fixtures.js
charlesstrange2: aha! it's an id, yes.

fixtures/fixtures.js
katjam: But technically speaking - according to our data - they are just trader objects. But I guess if you are just implying that this is a list of the traders attached to this user which has already been filtered by the API then maybe it makes sense.

Not going to fight you on it.

katjam:
@charlesstrange2 Are you also going to fix these mocks - so they can be used? I think Steph has hardcoded some stuff that breaks the login flow.

katjam (approved):
You deserve an ice cream! :icecream: :shipit:

PR #57, opened 2017-05-24
0.2/143 ui identifier logged in

gingerbreadpup:
Some static content here in the new profile bar, will need to hook it up later. Still a work in progress but maybe getting towards something now.

:sleeping:

### DIFF
--- a/src/components/Profile.vue
+++ b/src/components/Profile.vue
@@ -2,7 +2,7 @@
 
     <div class="profile-bar">
         <div><strong>2</strong> vouchers added</div>
-        <div>Logged in as <strong>P &amp; J Millward</strong></div>
+        <div>Logged in as <router-link v-bind:to="'/account'">P &amp; J Millward</router-link></div>
     </div>
 
 </template>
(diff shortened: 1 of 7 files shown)


src/components/Profile.vue
```diff
@@ -0,0 +1,14 @@
+<template id="profile">
+
+    <div class="profile-bar">
+        <div><strong>2</strong> vouchers added</div>
+        <div>Logged in as <strong>P &amp; J Millward</strong></div>
```
charlesstrange2: Worth adding an clicklink to accounts page on that name?

src/components/Toggle.vue
```diff
@@ -0,0 +1,15 @@
+<template id="toggle">
```
charlesstrange2: Toggle? toolbar perhaps?

charlesstrange2 (approved):
works, but just have some nomenclature ideas.
:sheep::it:
Ship it

src/components/Toggle.vue
gingerbreadpup: Yeah, that works!

PR #58, opened 2017-05-25
0.2/184 voucher count in navbar

gingerbreadpup:


tests/pages/tap.test.js
```diff
@@ -63,7 +63,7 @@ test('Page displays number of recorded vouchers', async t => {
         .typeText('#userPassword', 'secretpass')
         .pressKey('enter')
     ;
-    const voucherCount = await el('main#tap form+div').innerText;
+    const voucherCount = await el('#app > div.wrapper > div.profile-bar > div:nth-child(1)').innerText;
```
katjam, line 66: That's pretty specific.... I am never sure if less is more in testing... in case some of the markup around it changes.

katjam (approved):
Fine.

PR #60, opened 2017-05-26
0.2/177 user route is triggered if a user doesn t have a selected trader 2

charlesstrange2:
Also includes https://trello.com/c/ysCxzjqe/127-add-ui-element-to-make-clear-which-trader-working-on-behalf-of


src/components/Profile.vue
```diff
@@ -2,13 +2,22 @@
...
 export default {
-    name: 'profile'
+    name: 'profile',
+    data() {
+        return {
+            selectedTrader: Store.trader
+        }
+    },
+    created : function() {
+        console.log(this.selectedTrader);
```
katjam: Not saying anything.

src/main.js
```diff
@@ -42,20 +42,30 @@ const router = new VueRouter({
 
 // Route Guard rules for directing users
```
katjam, line 43: Just to clarify for me... what's strangers vs friends only

src/pages/User.vue
```diff
@@ -7,28 +7,20 @@
 
             <div class="content narrow">
 
-                <h1>Choose a trader account:</h1>
+                <form id="frmChooseTrader" v-on:submit.prevent >
 
-                <div class="form-group">
+                <h1>Choose a trader to manage</h1>
```
katjam: It feels like the page title should not be in the form.

I think form should only contain inputs, buttons etc - not other markup.

src/store.js
```diff
@@ -1,11 +1,13 @@
...
-    user: {"id": 1},
+    user: {
+        id : 1,
+        traders : []
+    },
     trader: {
-        "id": 1,
-        pendedVouchers: []
+        id: null,
+        pendedVouchers: [],
```
katjam: remove , ?

katjam (changes requested):
Just some small stuff.

src/main.js
charlesstrange2: strangers = pages only logged out people can use
friends = pages that only logged in people can use

src/pages/User.vue
charlesstrange2: Will move

src/store.js
charlesstrange2: oke

src/main.js
katjam: That's what I thought. Do we have strangers pages - besides login?

katjam (approved):
:ship: Once we get the tests passing. :+1:

PR #65, opened 2017-05-30
Correct API account history endpoint

gingerbreadpup:


charlesstrange2 (approved):
aha! good find!
:sheep::it:

PR #66, opened 2017-05-30
UI fixes

gingerbreadpup:
Making the UI less confusing. I don't remember when we decided to add the get payment button to the tap page, but it was pretty confusing so I've removed it, and have renamed 'Queued Vouchers' to 'Get Payment' in the header. Also I've renamed the main button to 'Submit code'.

### DIFF
--- a/src/pages/Tap.vue
+++ b/src/pages/Tap.vue
@@ -28,10 +28,7 @@
                       >
                     </div>
 
-                    <div class="two-buttons">
-                        <button v-on:click="record" id="submitVoucher" class="left" v-bind:class="{ cta : recVouchers[0] && recVouchers[0].length === 0 }">Add <span v-if="recVouchers[0] && recVouchers[0].length > 0">another </span><span v-else>voucher</span></button>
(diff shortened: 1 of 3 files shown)


src/pages/Tap.vue
```diff
@@ -28,10 +28,7 @@
                       >
                     </div>
 
-                    <div class="two-buttons">
-                        <button v-on:click="record" id="submitVoucher" class="left" v-bind:class="{ cta : recVouchers[0] && recVouchers[0].length === 0 }">Add <span v-if="recVouchers[0] && recVouchers[0].length > 0">another </span><span v-else>voucher</span></button>
-                        <button class="right" v-if="recVouchers[0] && recVouchers[0].length > 0">Get payment</button>
-                    </div>
+                    <button v-on:click="record" id="submitVoucher" class="left cta">Submit code</button>
```
katjam: Do we need to add the validation and dynamic class back in to the button - or does the function take care of it?

gingerbreadpup: Nope, we don't need any of it now.

charlesstrange2 (approved):
looks ok
:shipit:

PR #67, opened 2017-05-30
0.2/142 app logout change user facility

charlesstrange2:


gingerbreadpup (approved):
:sheep: :it:

src/store.js
```diff
@@ -15,6 +15,21 @@ var store = {
...
+    // TODO : I'm thinking we need some user/trader objects that manage this.
+    this.user = {
+        id: null,
+        traders : []
+    };
+    this.trader = {
+        id : null,
+        pendedVouchers : []
+    };
+    this.vouchers = this.vouchers.splice(0, this.vouchers);
```
katjam, line 29: Is this intended to clear the array? What is the value of this.vouchers at this point?

katjam (approved):
Apart from my one query... looks ok. :shipit:

src/store.js
charlesstrange2: at this point it could contain a voucher that someone was going to submit, but somehow managed not to. :-/ the slice business just zeros it without upsetting vue's observers, which, to be fair i dont care much about at this point in the spa. I'm quite keen to convert to vuex next sprint!

katjam: OK - keen not to over complicate - which is why my instinct was this.vouchers = []

So it does what it says in the name 'resetStore'

But happy with your judgement on this.

PR #68, opened 2017-05-30
Add error message for login page

gingerbreadpup:
I was going to add an error message for each login field, but after checking the OWASP docs and also seeing how it was done on MOMO, I decided to just have one error message display. Rather than telling the user which field was incorrect, it just says that the username and password combination they entered wasn't recognised.

### DIFF
--- a/src/pages/Login.vue
+++ b/src/pages/Login.vue
@@ -7,7 +7,7 @@
 
                 <h1>Log In</h1>
 
-                <div class="message">The username and password combination entered was not recognised. Please check your details and try again.</div>
+                <div class="message">{{ errorMessage }}</div>
 
                 <div>
                     <form id="loginForm" v-on:submit.prevent="onLogin">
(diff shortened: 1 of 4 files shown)


src/pages/Login.vue
```diff
@@ -7,15 +7,15 @@
 
                 <h1>Log In</h1>
 
+                <div class="message">The username and password combination entered was not recognised. Please check your details and try again.</div>
```
katjam: Fine to start but we probably need these to be the messages returned by the API. Are there any there?

src/pages/Login.vue
```diff
@@ -28,17 +28,21 @@
...
-
     export default {
         name: 'login',
         data: function () {
             return {
                 username: null,
                 password: null,
-                remember: true
+                remember: true,
+                errorMessage : Store.errors[0]
```
charlesstrange2: oddly, i dont think that [0] is necessary here. unless you *really are* watching the first error message in the array?

charlesstrange2 (changes requested):
see if you really need it to be store.errors[0]

charlesstrange2 (approved):
:sheep::it:

PR #70, opened 2017-05-30
0.2/146 initial tests for market frontend

alexalala:
https://trello.com/c/AxT3VNrm/146-initial-tests-for-market-frontend

all pass as can be seen on the trello card! 

Some functionality has probably been added since I started these so I'll write tests as and when they are needed.

🐵


alexalala:
@charlesstrange2 all ready for you to check now!!!

charlesstrange2 (approved):
My test cafe isn't great, but i can't see any problems in this - merge after Katja has reviewed/approved

tests/pages/account.test.js
```diff
@@ -0,0 +1,77 @@
+import { expect } from 'chai';
+import { Selector } from 'testcafe';
+
+const el = Selector(selector => document.querySelector(selector));
+
+const url = 'http://localhost:8081';
```
katjam: I would put this as http://localhost:8081/account

I know you have to log in but this will redirect you there. And will make it easier to see which page is being tested.

tests/pages/account.test.js
```diff
@@ -0,0 +1,77 @@
...
+        .typeText('#userPassword', 'secretpass')
+        .pressKey('enter')
+        .click("#radio-0")
+        .pressKey('enter')
+    ;
+    const accountNavButton = await el('a[href*="/account"]');
+
+    await t
+        .click(accountNavButton)
+    ;
```
katjam: If you use url /account you can eliminate lines 19-23

tests/pages/account.test.js
```diff
@@ -0,0 +1,77 @@
...
```
katjam: Won't need these if change url

katjam (commented):
initial thoughts.

tests/pages/account.test.js
alexalala: ahhh yes good shout.. I'll repeat on all the others as well

alexalala:
@katjam ready for round 2 of checking!

tests/pages/login.test.js
```diff
@@ -0,0 +1,57 @@
...
+
+test('Header text is correct', async t => {
+    const header = await el('.content h1').innerText;
+    expect(header).to.equal('Log In');
+});
+
+test('Must fill both input boxes to log in successfully', async t => {
+    await t
+        .typeText('#userName', 'email@example.com')
+        .pressKey('enter')
```
katjam: I think you might be asserting the wrong fail in firefox.

katjam (changes requested):
Awesome apart from that one... enter.

:lollipop:

katjam (approved):
:hand: go for it!

alexalala:
🙌

PR #71, opened 2017-05-30
Add better font for Firefox, add top margin on tap page

gingerbreadpup:
Changing to Roboto Slab regular, because bold looks ridiculous on Firefox. Huge difference between Chrome and Firefox.

Also adding a bit of margin to the top of the tap page so it's not all squashed, with the extra icons at the top right corner.


katjam (approved):
:shipit: If it looks good - fine by me!

PR #73, opened 2017-05-30
0.2/166 voucher count

gingerbreadpup:
Now if you add a voucher, the voucher count in the profile bar should fade out and in when it updates the number. It also does this when you switch trader, so you can see the effect by doing that (which is easier than trying to find a voucher you haven't put in already...)


charlesstrange2 (approved):
Looks ok, it'll be awesome!
:sheep::it:

PR #74, opened 2017-05-30
Improve UI, add logout button in profile bar, add change trader link

gingerbreadpup:
Some improvements to help navigation...

Added a logout button to the profile bar
Added a 'change trader' link
Made the username link to the account page
Moved the voucher count into the tap/scan/upload pages
Fix Firefox issue - the textbox was too wide, there was a horizontal scrollbar
Fix input icons position

### DIFF
--- a/src/components/Profile.vue
+++ b/src/components/Profile.vue
@@ -1,7 +1,7 @@
 <template id="profile">
 
     <div class="profile-bar">
-        <div><router-link v-bind:to="'/account'"><strong>{{ selectedTrader.name }}</strong></router-link> <router-link v-bind:to="'/user'">(Change trader)</router-link></div>
+        <div><strong>{{ selectedTrader.name }}</strong> <router-link v-bind:to="'/user'">(Change trader)</router-link></div>
         <a v-on:click="onLogout">Log out</a>
     </div>
 
--- a/src/components/Profile.vue
+++ b/src/components/Profile.vue
@@ -1,10 +1,8 @@
 <template id="profile">
 
     <div class="profile-bar">
-        <div>Managing <router-link v-bind:to="'/user'"><strong>{{ selectedTrader.name }}</strong></router-link></div>
-        <transition name="fade" mode="out-in">
-            <div :key="voucherCount" v-if="vouchersAdded"><router-link v-bind:to="'/payment'"><strong>{{ voucherCount }}</strong> voucher<span v-if="voucherCount > 1">s</span> waiting</router-link></div>
-        </transition>
+        <div><router-link v-bind:to="'/account'"><strong>{{ selectedTrader.name }}</strong></router-link> <router-link v-bind:to="'/user'">(Change trader)</router-link></div>
+        <a v-on:click="onLogout">Log out</a>
     </div>
 
 </template>
@@ -15,20 +13,13 @@ export default {
     name: 'profile',
     data: function() {
         return {
-            selectedTrader : Store.trader,
-            recVouchers : Store.recVouchers,
-            voucherCount : 0
+            selectedTrader : Store.trader
         }
     },
-    mounted: function() {
-        Store.getRecVouchers();
-    },
-    computed: {
-        vouchersAdded: function() {
-            if (this.recVouchers[0] && this.recVouchers[0].length > 0) {
-              this.voucherCount = this.recVouchers[0].length;
-              return true;
-            }
+    methods: {
+        onLogout : function() {
+            Store.unAuthenticate();
+            this.$router.push('/login');
         }
     }
 }
--- a/src/components/Toolbar.vue
+++ b/src/components/Toolbar.vue
@@ -2,9 +2,9 @@
 
     <div class="toolbar">
         <transition name="fade" mode="out-in">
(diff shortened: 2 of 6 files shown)


src/components/Profile.vue
```diff
@@ -1,10 +1,8 @@
 <template id="profile">
 
     <div class="profile-bar">
-        <div>Managing <router-link v-bind:to="'/user'"><strong>{{ selectedTrader.name }}</strong></router-link></div>
-        <transition name="fade" mode="out-in">
-            <div :key="voucherCount" v-if="vouchersAdded"><router-link v-bind:to="'/payment'"><strong>{{ voucherCount }}</strong> voucher<span v-if="voucherCount > 1">s</span> waiting</router-link></div>
-        </transition>
+        <div><router-link v-bind:to="'/account'"><strong>{{ selectedTrader.name }}</strong></router-link> <router-link v-bind:to="'/user'">(Change trader)</router-link></div>
+        <a v-on:click="onLogout">Log out</a>
     </div>
```
charlesstrange2: There's a lot of UI stuff in that corner - on a small screen is there a danger of fat-finger syndrome?
also, should logout be a linky or a styled button/icon?

src/components/Toolbar.vue
```diff
@@ -1,15 +1,38 @@
...
-        <router-link v-bind:to="'/'"><i class="fa fa-keyboard-o" aria-hidden="true"></i></router-link>
-        <router-link v-bind:to="'/scan'"><i class="fa fa-camera" aria-hidden="true"></i></router-link>
-        <router-link v-bind:to="'/upload'"><i class="fa fa-upload" aria-hidden="true"></i></router-link>
+        <transition name="fade" mode="out-in">
+            <div :key="voucherCount" v-if="vouchersAdded" class="count"><router-link v-bind:to="'/payment'"><strong>{{ voucherCount }}</strong> voucher<span v-if="voucherCount > 1">s</span> waiting</router-link></div>
+        </transition>
+        <div class="input-icons">
+            <router-link v-bind:to="'/'"><i class="fa fa-keyboard-o" aria-hidden="true"></i></router-link>
+            <router-link v-bind:to="'/scan'"><i class="fa fa-camera" aria-hidden="true"></i></router-link>
+            <router-link v-bind:to="'/upload'"><i class="fa fa-upload" aria-hidden="true"></i></router-link>
```
charlesstrange2, line 10: scan and upload are blue, tap is grey. Muted or pastel colours (like grey) often denote "deactivated"; consider 
 swapping, so that the active tool is in the richer colour.

src/components/Profile.vue
```diff
@@ -1,10 +1,8 @@
 <template id="profile">
 
     <div class="profile-bar">
-        <div>Managing <router-link v-bind:to="'/user'"><strong>{{ selectedTrader.name }}</strong></router-link></div>
-        <transition name="fade" mode="out-in">
-            <div :key="voucherCount" v-if="vouchersAdded"><router-link v-bind:to="'/payment'"><strong>{{ voucherCount }}</strong> voucher<span v-if="voucherCount > 1">s</span> waiting</router-link></div>
-        </transition>
+        <div><router-link v-bind:to="'/account'"><strong>{{ selectedTrader.name }}</strong></router-link> <router-link v-bind:to="'/user'">(Change trader)</router-link></div>
```
charlesstrange2: - (change trader) is good, but it's a function that the trader name used to have. I see why you've moved it to make it an explicit option, it was otherwise not obvious how to swap traders.

- The trader name still doesn't advertise it's active or where it goes; hidden discoverable? given the My Account page alreday exists and is advertised, why link the name in this way?

charlesstrange2 (changes requested):

src/components/Profile.vue
gingerbreadpup: No, login doesn't have to be a button, or have an icon necessarily - it has an underline when you hover. I think any more icons and we'll risk going overboard with them.

gingerbreadpup: I realise that the hover won't be something you see when using the app on a tablet or phone. We can maybe think about improvements to the login button in the next sprint, I think the main thing here is that it's been moved to a place where you can easily access it.

charlesstrange2:
can you hover on a mobile device?

charlesstrange2:
ok, fair point. we'll do rock-paper-scissors another day about it :-)

src/components/Profile.vue
gingerbreadpup: I don't think it made sense to have the trader name serving as the link to the change trader page, which is why I added the Change Trader link.

You're right, we don't necessarily need to have the trader name linking to the account page, but since there's no other place I can think of that it should link to, I don't think there's any harm in having another link to the account page.

I can remove the link though if you like and just leave it as the trader name. Let me know what you think!

src/components/Profile.vue
gingerbreadpup: There is definitely a lot of UI stuff in that corner - I think maybe it's something to look at in the next sprint. I'm happy with where everything is right now as that's how it is in the paper prototype.

charlesstrange2 (approved):
:shipit:

PR #76, opened 2017-05-31
0.2/blue links

gingerbreadpup:
One time. This one actually works because I've tested it on Alex's computer and her blue links have gone. Hurrah!


charlesstrange2 (approved):
:shipit:

PR #77, opened 2017-05-31
Small renaming tweaks, and some more realistic payment instructions

gingerbreadpup:


charlesstrange2 (approved):
do we need that extra commented out button?
:shipit:

PR #78, opened 2017-05-31
0.2/pulse

gingerbreadpup:
Took a bit of fiddling but got the pulse animation to work as a Vue transition. So now it pulses instead of fades, it grabs your attention a bit better.


charlesstrange2 (approved):
cool
:shipit:

PR #81, opened 2017-06-02
Make logo a link if you're past logged in and have selected a trader

gingerbreadpup:


katjam (approved):
Seems like an OK solution. :)

:seedling:

PR #82, opened 2017-06-02
Converted incoming response object-of-vouchers into an array of vouch…

charlesstrange2:
…ers.

Not sure this *fixes* the problem as it's described as transient. It will however fix the "e.sort" problem - you can't sort an object-of-objects, but you can sort an array-of-objects!


gingerbreadpup (approved):
:chipmunk: :it:

PR #85, opened 2017-06-02
0.2/218 clear failed voucher codes on submit

charlesstrange2:


src/pages/Payment.vue
```diff
@@ -84,11 +84,11 @@ export default {
         onRequestPayment() {
             Store.pendRecVouchers(
                 // on Success, route to /account
-                function() {
+                function(response) {
```
katjam, line 87: You need?

src/pages/Payment.vue
```diff
@@ -84,11 +84,11 @@ export default {
...
                     this.$router.push('/account');
                 }.bind(this),
                 // on Failure... hook for an alert?
-                function() {
+                function(error) {
```
katjam, line 91: And this?

src/pages/Tap.vue
```diff
@@ -60,12 +63,19 @@ export default {
             if (this.voucherCode !== null && this.voucherCode.length > 0) {
                 Store.addVoucherCode(this.sponsorCode.toUpperCase()+this.voucherCode,
                     // Success function
-                    function() {
+                    function(response) {
+                            // Add error message for invalid and fail codes.
```
katjam: double indent - but ok otherwise

src/store.js
```diff
@@ -186,10 +186,11 @@ store.transitionVouchers = function (transition, vouchers, success, failure) {
         'vouchers': vouchers
     };
     this.netMgr.apiPost('vouchers', postData,
-        function () {
-            if (success) {success()}
-        }, function() {
-            if (failure) {failure()}
+        function (response) {
+            if (success) {success(response);}
+        }.bind(this),
```
katjam: don't need the bind anymore, nor the ; in the inline ifs.

katjam (approved):
Few little syntax tidys then :shipit:

src/pages/Payment.vue
charlesstrange2: I think so, will JS throw a fit if it's passed a variable in a function when it's not expecting it?

src/pages/Payment.vue
charlesstrange2: ditto above - transition is now feeding response and error objects to success/failure callbacks.

src/store.js
charlesstrange2: okedoke

PR #87, opened 2017-06-05
Fix error message transition

gingerbreadpup:
Wondered why my error message transitions weren't working, then noticed the > was missing. :cake:


katjam (approved):
lol :seedling:

PR #90, opened 2017-06-05
Altered the endpoint target to include the correct status query param

charlesstrange2:
Just needed the query string updating to be more selective.


katjam (approved):
:hammer:

PR #91, opened 2017-06-05
Removed sorting function from store.getRecVouchers

charlesstrange2:
We trust the server to pre-sort for us.


katjam (commented):
:As long as we trust the server. :+1:

katjam (approved):

PR #93, opened 2017-06-05
0.2/245 download buttons on account screen

gingerbreadpup:


src/pages/Account.vue
```diff
@@ -25,6 +26,7 @@
                                 <div> {{ payment.pended_on }}</div>
                                 <div> {{ payment.vouchers.length }}</div>
                                 <div class="amount">£{{ payment.vouchers.length }}</div>
```
katjam: pound sign should be encoded

charlesstrange2 (approved):

src/pages/Account.vue
```diff
@@ -25,6 +26,7 @@
...
+                                <div class="download"><i class="fa fa-download" aria-hidden="true" title="Download"></i></div>
```
charlesstrange2, line 29: When it's running we'll need to insert data based on payment.pended_on in here for vue wiring.

PR #101, opened 2017-06-12
Remove download button from payment page, other style tweaks

gingerbreadpup:
Also I've added better date descriptions on the account page. This is part of another card on Trello so I should have done it in a separate branch I guess - but I will link this PR in that card.

### DIFF
--- a/src/pages/Account.vue
+++ b/src/pages/Account.vue
@@ -11,7 +11,7 @@
                     <div class="tab thead">
                         <label>
                             <div class="row">
-                                <div>Date</div>
+                                <div>Request date</div>
                                 <div class="total"></div>
                                 <div>Amount</div>
                                 <div></div>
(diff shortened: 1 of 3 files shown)


src/pages/Account.vue
```diff
@@ -25,16 +25,16 @@
                             <div class="row">
                                 <div> {{ payment.pended_on }}</div>
                                 <div> {{ payment.vouchers.length }}</div>
-                                <div class="amount">£{{ payment.vouchers.length }}</div>
-                                <div class="email"><i class="fa fa-envelope" aria-hidden="true" title="Send record to my email"></i></div>
+                                <div class="amount">&pound;{{ payment.vouchers.length }}</div>
```
katjam, line 28: I had a look at this the other day and according to some sources using the character is fine as long as you have utf-8 declared in the doc type. I prefer the encoded ones anyway - but would be good to know what the standard is today.

katjam (approved):
:field_hockey: Looking great!

charlesstrange2 (approved):
Looks good, worth investigating the UTF-8 thing though.

PR #103, opened 2017-06-13
1.0/96 spinner timeout

gingerbreadpup:
Not passing tests on here, but is passing them all fine on my laptop. Will give it another go tomorrow...

PR #104, opened 2017-06-14
1.0/more specific errors

alexalala:
https://trello.com/c/sgaTPp9d/114-more-specific-errors-on-failed-or-invalid-voucher-submission
https://trello.com/c/tZUbQoFz/265-front-end-test-for-voucher-submit

killing 2 cards with 1 PR 👍 

🐦


gingerbreadpup (dismissed):
Looks good to me :shipit:

src/pages/Scan.vue
```diff
@@ -76,9 +76,12 @@ export default {
                         // Add error message for invalid and fail codes.
                         if (
                             response.data.invalid.length > 0
-                            || response.data.fail.length > 0
                         ) {
-                            this.errorMessage = "The code you entered is not valid. Please try again.";
+                            this.errorMessage = "The voucher code you entered is not valid. Please try again.";
```
katjam, line 80: When this goes in... Be aware that this has changed in the https://github.com/neontribe/ARCVMarket/pull/103.

fixtures/fixtures.js
```diff
@@ -90,12 +102,27 @@ Fixtures.apply = function (mock) {
...
+        });
         var data = {
-            "success": JSON.parse(request.data).vouchers,
-            "fail": [],
-            "invalid": []
+            success,
+            fail,
+            invalid
         };
         console.log(data);
```
katjam, line 127: Probably don't want this in here.

fixtures/fixtures.js
```diff
@@ -90,12 +102,27 @@ Fixtures.apply = function (mock) {
     mock.onGet(/\/traders\/2\/vouchers/).reply(200, this.traderVouchers["2"]);
 
     mock.onPost('/vouchers').reply(function (request) {
```
katjam, line 104: Shouldn't the reply be a fixtured voucherStatus like the api returns:

```
'success': ['RVP12345566'],
'fail: ['FAL12345667'],
'invalid' : ['INV11']
```

fixtures/fixtures.js
```diff
@@ -90,12 +102,27 @@ Fixtures.apply = function (mock) {
...
-        // returns a success regardless!
-        // TODO : better emulation of server side validated responses;
+        const requestPayload = JSON.parse(request.data);
+
+        const success = [];
+        const fail = [];
+        const invalid = [];
+
+        requestPayload.vouchers.forEach((voucherNumber) => {
```
katjam, line 111: where are you getting the voucher numbers from? If we are on fixtures - I think we just use the fixtures - not actually take input. @charlesstrange2 What do you think?

fixtures/fixtures.js
```diff
@@ -43,6 +43,18 @@ var Fixtures = {
             {"code": "SOL00000012", "updated_at": "17-05-2017 14:46.15"},
         ]
     },
+    vouchersStatus: {
```
katjam, line 46: These aren't being used, are they?. I think you need to set up as I described below so it matches what the api does - rather than being a list of voucher objects. Should be the status returned, not the vouchers.

katjam (changes requested):
Requires some input from @charlesstrange2 also - but simpler seems better to me in this case.

fixtures/fixtures.js
charlesstrange2: traderVouchers["1"] isn't being used, true; could be pruned.

fixtures/fixtures.js
charlesstrange2: It's GET the list of vouchers that are already approved; that should come back as an object list of correctly assigned vouchers.. 
The format above is part of the POST response, which tells you the result of the action you've done (i.e. "please tell me if this list of 1(or more) vouchers were progressed correctly")

fixtures/fixtures.js
charlesstrange2: Agreed - fixtures represent a given app-state. You're writing a loose emulation of the serverside in here, which is  more than we need (even though it is more flexible).

charlesstrange2 (commented):
rewrite to fixtures rather than function

charlesstrange2 (changes requested):
rewrite to fixtures rather than function

katjam:
@alexalala 

- Clear out unused user, 
- add 3 voucherStatus fixtures keyed by fail, success, invalid
- make a switch in your mock responder - for all 3 cases plus a default of success (so any old voucher will be successful)
- response looks like: {"success":["RVP12345560"],"fail":[],"invalid":[]}

PR #105, opened 2017-06-14
1.0/284 create a pwa manifest

charlesstrange2:


src/manifest.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name" : "ARC Voucher Recorder",
```
katjam: Do we need to [xXx] any of this copy?

katjam (approved):
Looks great unless we need to run the copy past anyone.

:orange_book: :panda_face:

src/manifest.json
gingerbreadpup: Rose Voucher Recorder? ARC won't mean much to the traders.

gingerbreadpup (approved):
splendid! :shipit: :beach_umbrella:

PR #106, opened 2017-06-14
Set ff version to v53

ob6160:


.travis.yml
```diff
@@ -4,7 +4,7 @@ node_js:
 sudo: required
 dist: trusty
 addons:
-  firefox: latest
+  firefox: 53.0
```
katjam: Ace. Maybe add comment to revisit in a bit when test cafe updates - because we don't want to be here for long.

katjam (approved):
+ Either a comment or a card to revisit.

.travis.yml
ob6160: Agreed, will do.

PR #110, opened 2017-06-15
1.0/269 email voucher history

katjam:
Hack by me to get the button working. Please tidy before or after merge.

### DIFF
--- a/src/pages/Account.vue
+++ b/src/pages/Account.vue
@@ -51,7 +51,7 @@
 
                 </div>
 
-                <button>Email payment history</button>
+                <button id="requestVoucherHistoryEmail" v-on:click="onRequestVoucherHistoryEmail(null)">Email payment history</button>
 
             </div>
 
@@ -61,6 +61,7 @@
 
 <script>
     import Store from '../store.js';
+    import NetMgr from '../services/netMgr.js';
     export default {
         name: 'account',
         data() {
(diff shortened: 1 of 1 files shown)


gingerbreadpup (approved):
:shipit:

src/pages/Account.vue
```diff
@@ -71,6 +72,18 @@
         created: function () {
 
         },
+        methods: {
+            onRequestVoucherHistoryEmail(submittedOn = null) {
+                NetMgr.apiPost('/traders/' + Store.trader.id + '/voucher-history-email?submission_date=' + submittedOn,
+                    function (response) {
+                        if (success) { success(response) }
```
charlesstrange2, line 79: unneeded function doesn't take a callback

src/pages/Account.vue
```diff
@@ -71,6 +72,18 @@
...
+                    },
+                    function () {
+                        if (failure) { failure(error) }
```
charlesstrange2, line 82: unnneded, function doesnt take a callback

src/pages/Account.vue
```diff
@@ -71,6 +72,18 @@
         created: function () {
 
         },
+        methods: {
+            onRequestVoucherHistoryEmail(submittedOn = null) {
```
charlesstrange2, line 76: if it's an onclick handler, i'd expect submitteOn to be stuffed with an event?

charlesstrange2 (changes requested):
Is that not receiving an onclick event in that submittedOn variable?

katjam:
@charlesstrange2 or @gingerbreadpup 

If you can fix it great... All I know is that it currently works :) even if it is badly written.

The next step for me is to add the submitted variable to the little icons that ask for emails per submitted date -which is where that param will be used. I am using the null value as a default to mean 'all history' but the other buttons will submit a value.

PR #114, opened 2017-06-16
Accessibility fixes

gingerbreadpup:
Passes the accessibility checks on WAVE and Chrome Lighthouse now. Tried to find a better way to solve the problem and couldn't - but this way seems to be accepted as an okay way of doing it, even if it's not ideal - so hopefully this will be sufficient for now. :grimacing:

### DIFF
--- a/src/components/Masthead.vue
+++ b/src/components/Masthead.vue
@@ -4,7 +4,7 @@
 
         <header>
             <div class="logo" v-bind:class="{ login : $route.path==='/login' || $route.path==='/user' }">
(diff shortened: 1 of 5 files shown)


charlesstrange2 (approved):
The whole "hidden offscreen" thing smells a bit. Card for improvement

katjam:
@charlesstrange2 Hidden offscreen is a common and acceptable way of hiding stuff. (As far as I know)

katjam (approved):

charlesstrange2:
Yes, explored that with steph; make me a little bit sad, but it works, and
passes the accessibility check to get me my 100% in Lighthouse. :-)

katjam:
I've seen it done that way by many many people including Matt and Heydon -
so it can't be too terrible.

PR #117, opened 2017-06-19
prevent default  wont fire if keypress is backspace

alexalala:
https://trello.com/c/j9EoVMY8/238-bug-in-firefox-cannot-use-backspace-key-in-the-number-part-of-the-voucher-input-box

🐖

### DIFF
--- a/src/pages/Tap.vue
+++ b/src/pages/Tap.vue
@@ -162,7 +162,8 @@ export default {
             var rxNumber = /\d/;
             var char = this.getKeyCharCode(event);
             
-            if (event.keyCode != 8) {
+            //event.keycode 8 is backspace, dont want to prevent default
+            if (event.keyCode !== 8) {
                 if (char.match(rxNumber)) {
                     if (this.voucherCode.length < event.target.maxlength) {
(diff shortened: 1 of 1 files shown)


src/pages/Tap.vue
```diff
@@ -162,15 +162,17 @@ export default {
         onKeypressVoucherBox : function(event) {
             var rxNumber = /\d/;
             var char = this.getKeyCharCode(event);
-
-            if (char.match(rxNumber)) {
-                if (this.voucherCode.length < event.target.maxlength) {
-                    this.voucherCode += char;
+            
+            if (event.keyCode != 8) {
```
katjam: Comment that event.keycode 8 is backspace

and use !== to be typesafe?

katjam (changes requested):

src/pages/Tap.vue
alexalala: @katjam ok!

katjam (approved):
:label: Good stuff.

charlesstrange2 (approved):
Neat!
:sheep::it:

PR #118, opened 2017-06-19
Fix for accessibility label height and header shadow

gingerbreadpup:
Two tiny changes. One fixes something I put in for accessibility (label height above add voucher input), and the other adds a header shadow which I'd been meaning to put in for a while.

Still working on another branch to improve the queue list... but not finished yet.

### DIFF
--- a/src/sass/app.scss
+++ b/src/sass/app.scss
@@ -127,6 +127,8 @@ $fa-font-path:"../node_modules/font-awesome/fonts";
 
 .wrapper {
   position: relative;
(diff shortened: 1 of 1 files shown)


src/sass/app.scss
```diff
@@ -482,6 +484,10 @@ main > *:target ~ * {
   }
 }
 
+label#lblSponsorBox {
+  height: 0;
+}
+
```
katjam, line 490: Extra line. Don't mind if you keep it or not. :)

katjam (approved):
:wilted_flower: Hot!

PR #121, opened 2017-06-20
1.0/114 more specific errors

katjam:
Tests passing. We can make the voucher submit fixtures more dynamic in next sprint. [xXx] the errors because they are too long for @gingerbreadpup current design.

### DIFF
--- a/fixtures/fixtures.js
+++ b/fixtures/fixtures.js
@@ -10,17 +10,6 @@ var Fixtures = {
                 "created_at": "24-05-2017 14:19:22",
                 "updated_at": "24-05-2017 14:19:22",
                 "deleted_at": null
-            }
-        ],
-        "2": [
-            {
-                "id": 1,
-                "name": "Kristy Corntop",
-                "pic_url": null,
-                "market_id": 2,
-                "created_at": "24-05-2017 14:19:22",
-                "updated_at": "24-05-2017 14:19:22",
-                "deleted_at": null
             },
             {
                 "id": 2,
@@ -42,6 +31,14 @@ var Fixtures = {
             {"code": "SOL00000015", "updated_at": "17-05-2017 13:46.15"},
             {"code": "SOL00000012", "updated_at": "17-05-2017 14:46.15"},
         ]
+    },
+    voucherStatus: {
+        "success":
+            {"success":["RVP12345678"], "fail":[], "invalid":[]},
+        "fail":
+            {"success":[], "fail":["FAL11111111"], "invalid":[]},
+        "invalid": 
+            {"success":[], "fail":[], "invalid":["INV1"]},
     },
     traderVoucherHistory: {
         "1": [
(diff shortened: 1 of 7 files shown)


fixtures/fixtures.js
```diff
@@ -43,6 +32,14 @@ var Fixtures = {
             {"code": "SOL00000012", "updated_at": "17-05-2017 14:46.15"},
         ]
     },
+    voucherStatus: {
+        "success":
+            {"success":["RVP12345678"], "fail":[], "invalid":[]},
+        "fail":
+            {"success":[], "fail":["FAL11111111"], "invalid":[]},
+        "invalid": 
+            {"success":[], "fail":[], "invalid":["INV1"]},
```
alexalala, line 41: prob dont need this comma?

charlesstrange2 (approved):
looks good
:shipit:

src/pages/Scan.vue
```diff
@@ -85,10 +85,13 @@ export default {
...
-                            || response.data.fail.length > 0
                         ) {
                             this.showFail();
-                            this.errorMessage = "Please enter a valid code.";
+                            this.errorMessage = "The voucher code you entered is not valid. Please try again.";
+                        } else if (
+                            response.data.fail.length > 0
+                        ) {
+                            this.errorMessage = "The voucher code has you entered has previously been submitted. Please try again.";
                         } else {
```
alexalala: xXx would be good on this page too!

alexalala (commented):
tiny changes!

src/pages/Scan.vue
charlesstrange2: Ooh, good spot! Unless it's been Steph-Stamped as good to go, [xXx] is probably the rule.

fixtures/fixtures.js
katjam: Trailing commas are used for 2 reasons:

1- git changes are more accurate when a new item is added to the end of a list (you only see the new line, not the old last line with a comma and the new one.
2- less likely to forget to add one to the line above and cause compile errors when adding to the bottom of the list.

But you are right that is is not needed and the fixtures file has some with and some without.

src/pages/Scan.vue
katjam: Those are really the same error - and should use the same message, rather than be repeated in 2 places. Shouldn't the Scan page and the Tap page use the same submit methods anyway?

fixtures/fixtures.js
alexalala: @katjam Ah cool thanks for the explanation! Wondered why they were inconsistent in the fixtures..

alexalala (approved):
🦐  🇮🇹  looks great!

src/pages/Scan.vue
katjam: done. And added info to cards... 

https://trello.com/c/o9kY8iNK/332-add-message-component-to-display-success-and-error-messages 

https://trello.com/c/o9kY8iNK/332-add-message-component-to-display-success-and-error-messages

PR #122, opened 2017-06-20
source map set to false

alexalala:
Seems to work fine! Screenshots attached to trello card https://trello.com/c/InIGkayp/302-optimise-build-size-dont-include-buildmapjs-in-production


charlesstrange2 (approved):
Awesome bool flipping!
:dolphin:

PR #124, opened 2017-06-20
1.0/336 endless spinner

gingerbreadpup:
Shortened the error messages for now - since the copy is just temporary we might as well have something that fits (in case of any more demos!). Once we decide on the final copy, I'll address the issue of making sure there's enough space to display the messages (already part of another card in tech debt).

Added fix for the spinner button so if you try to submit with no input, you get an error message and the button goes red and resets back to purple.

### DIFF
--- a/src/pages/Tap.vue
+++ b/src/pages/Tap.vue
@@ -66,7 +66,6 @@ export default {
     },
     methods:  {
         onRecordVoucher: function(event) {
-
             //TODO: some proper validation
             if (this.voucherCode !== null && this.voucherCode.length > 0) {
                 this.startSpinner();
--- a/src/pages/Tap.vue
+++ b/src/pages/Tap.vue
@@ -82,7 +82,7 @@ export default {
                             response.data.fail.length > 0
                         ) {
                             this.showFail();
-                            this.errorMessage = "[xXx] Please enter an unused voucher code.";
+                            this.errorMessage = "[xXx] That voucher may have been used already.";
                         } else {
                             this.showValidate();
                             this.errorMessage = "";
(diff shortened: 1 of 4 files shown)


src/pages/Tap.vue
```diff
@@ -77,11 +77,12 @@ export default {
...
                         ) {
                             this.showFail();
-                            this.errorMessage = "[xXx]The voucher code you entered is not valid. Please try again.";
+                            this.errorMessage = "[xXx] Please enter a valid voucher code.";
                         } else if (
                             response.data.fail.length > 0
                         ) {
-                            this.errorMessage = "[xXx]The voucher code you entered has previously been submitted. Please try again.";
+                            this.showFail();
+                            this.errorMessage = "[xXx] Please enter an unused voucher code.";
```
charlesstrange2: Makes me feel, as a user, like i didi this *on purpose* and am a bad person.
"That voucher may have been used already."

tests/pages/tap.test.js
```diff
@@ -153,7 +153,7 @@ test('Correct error appears when I submit an invalid voucher', async t => {
         .click(submitVoucher)
     ;
     const errorMessage = await el('.message').innerText;
-    expect(errorMessage).to.contain('The voucher code you entered has previously been submitted. Please try again.');
+    expect(errorMessage).to.contain('Please enter an unused voucher code.');
```
charlesstrange2: Change to suit the other changes above.

charlesstrange2 (changes requested):
Code fine, not sure about the wording. A bit accusatory of the poor befuddled user people.

src/pages/Tap.vue
gingerbreadpup: Sure, we can put whatever you like - mainly just wanted it to be shorter until we have approved copy. Will make changes now.

charlesstrange2 (approved):
:sheep::it:

PR #126, opened 2017-06-21
1.0/341 bug offline plugin does not consistently reload fresh builds

charlesstrange2:
The [hash] parts on the new filenames change every build, *thus* when we push a new build, the appcache and service worker should notice!
There's now a console.log on startup that prints the date, short hash and branchname to the console.log - I'm hoping that should be enough for testers to check that the current build is in place.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -19,7 +19,7 @@ import Payment from './pages/Payment.vue';
 import Login from './pages/Login.vue';
 import User from './pages/User.vue';
 
-console.log(BUILDDATE + ' ' + VERSION + ' ' + BRANCH);
+console.info(BUILDDATE + ' ' + VERSION + ' ' + BRANCH);
 
 /*
 route access rules
--- a/src/main.js
+++ b/src/main.js
@@ -19,7 +19,7 @@ import Payment from './pages/Payment.vue';
 import Login from './pages/Login.vue';
 import User from './pages/User.vue';
 
-console.info(BUILDDATE + ' ' + VERSION + ' ' + BRANCH);
+console.info(BUILDDATE + '\n' + BRANCH + '\n' + VERSION);
 
 /*
 route access rules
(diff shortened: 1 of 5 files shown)


src/main.js
```diff
@@ -19,6 +19,8 @@ import Payment from './pages/Payment.vue';
 import Login from './pages/Login.vue';
 import User from './pages/User.vue';
 
+console.log(BUILDDATE + ' ' + VERSION + ' ' + BRANCH);
+
```
gingerbreadpup: console.log left in there.

charlesstrange2: Yup! deliberately; that's a bit that a tester can check to make sure they're on the right build on thier device (if it's hooked up to a desktop with chrome); Unless there is a better place for a version number in the UI ?

ob6160: As it's just an informational message we could use console.info(https://developer.mozilla.org/en-US/docs/Web/API/Console/info) :innocent:

gingerbreadpup: lovely!

ob6160: Also it could possibly go in the main app mounted method: https://github.com/neontribe/ARCVMarket/pull/126/files/51d029d102133af4a1c276a49e5671a1a503a8e4..60d07664adab306cb2a8588a82be47d266af50ae#diff-1c90ff38a08209f9ebd4d05d1e43358eL98

src/pages/Login.vue
```diff
@@ -24,7 +24,7 @@
         </main>
 
         <div class="footer">
-            <span id="build">{{ buildmsg }}</span>
+            <span id="build">Commit : {{ commitmsg }}</span>
         </div>
```
gingerbreadpup: might as well be semantic and use` <footer>` instead of `<div class="footer">`

gingerbreadpup (changes requested):
Use the footer tag

charlesstrange2:
ah, yes. true.

src/main.js
ob6160: perhaps some newlines? console.info(BUILDDATE + "\n" + VERSION + "\n" + BRANCH);

ob6160 (approved):
:1st_place_medal:

gingerbreadpup (approved):
:shipit:

PR #128, opened 2017-06-21
1.0/redo auth service local storage

ob6160:
It's not perfect but it will do the job for now.

I've made two preliminary tests, @charlesstrange2 some more need to be made, we should figure out which other scenarios we need to test for.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -121,6 +121,5 @@ var vm = new Vue({
  */
 EventBus.$on('NetMgr.logout', (err) => {
     Store.resetStore();
-
     router.push('login');
 });
--- a/src/main.js
+++ b/src/main.js
@@ -113,7 +113,13 @@ var vm = new Vue({
     router
 }).$mount('#app'); // Mount the router on the app
 
-EventBus.$on('NetMgr.logout', () => {
+/**
+ * Reset all stored information and redirect the user back to the login page when 'NetMgr.logout' is fired.
+ *
+ * @param err
+ *   Logout reason.
+ */
+EventBus.$on('NetMgr.logout', (err) => {
     Store.resetStore();
 
     router.push('login');
--- a/src/main.js
+++ b/src/main.js
@@ -114,5 +114,7 @@ var vm = new Vue({
 }).$mount('#app'); // Mount the router on the app
 
 EventBus.$on('NetMgr.logout', () => {
+    Store.resetStore();
+
     router.push('login');
 });
--- a/src/main.js
+++ b/src/main.js
@@ -20,11 +20,11 @@ import Login from './pages/Login.vue';
 import User from './pages/User.vue';
 
 /*
-route access rules
-auth -> true, user MUST be auth'd - friends only
-auth -> false, user MUST NOT be auth'd - stranger's only
-auth -> undefined, auth not important - public
-*/
+ route access rules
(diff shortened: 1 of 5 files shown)


src/main.js
```diff
@@ -106,3 +112,9 @@ var vm = new Vue({
     // Pass in the router to the Vue instance
     router
 }).$mount('#app'); // Mount the router on the app
+
+EventBus.$on('NetMgr.logout', () => {
+    Store.resetStore();
+
```
charlesstrange2: Spacing

src/services/netMgr.js
```diff
@@ -1,17 +1,26 @@
+import Vue from 'vue';
```
charlesstrange2: Why include this? i don't think we use it in this file.

src/services/netMgr.js
```diff
@@ -35,32 +44,36 @@ var NetMgr = {
...
  *
  * @param tokenData
  */
 NetMgr.setToken = function (tokenData) {
     this.token = tokenData;
+
     if (this.token) {
         this.token.requestTime = Math.floor(Date.now() / 1000);
         this.axiosInstance.defaults.headers.common['Authorization'] = 'Bearer ' + this.token.access_token;
+
```
charlesstrange2, line 53: Spacing?

charlesstrange2 (changes requested):
Awesome!
See changes though

src/services/netMgr.js
ob6160: whoops I left that in, good catch

src/services/netMgr.js
ob6160: I quite like the space here as those two are variable sets and the last one is a method call. Don't mind either way - your call.

src/services/netMgr.js
```diff
@@ -2,16 +2,24 @@ import Config from '../config.js';
...
-- Authentication
+ NetMgr is a place to put all the
+ - AJAX axios stuff;
+ - Mock binding
+ - Authentication
 
-Though, in theory it could have been split into a series of modules.
+ Though, in theory it could have been split into a series of modules.
+
+ - event bus
```
katjam: Are these todos? It't really clear what this list is.

src/services/netMgr.js
katjam: I am with @charlesstrange2 on this one. Though not much of a leg to stand on - we have been sort of following 'standards' and the rest of the code isn't so gapp-y.

Happy to leave it and wait until we get a standards set agreed.

src/services/netMgr.js
```diff
@@ -160,37 +172,84 @@ NetMgr.mockOff = function () {
...
+    NetMgr.mockOn();
 }
 
+/**
+ * Sets the internal token variable to the result of the NetMgr.getTokenFromLocalStorage function.
+ */
+NetMgr.setTokenFromLocalStorage = function() {
+    let parsedLocalToken = this.getTokenFromLocalStorage();
+
+    if(parsedLocalToken) this.setToken(parsedLocalToken);
```
katjam: Up to @charlesstrange2 - another standards issue... inline if statement.

katjam (approved):
@charlesstrange2 I had a few minor comments - but happy for you to merge this in. Looks good.

charlesstrange2 (approved):

PR #131, opened 2017-06-22
Add new status to submit code button and message

gingerbreadpup:
This code currently doesn't work because it just needs a condition for the if statement, which should check to see if the app is offline, and then it will call the showQueued() function and will make the submit code button orange.

### DIFF
--- a/src/pages/Account.vue
+++ b/src/pages/Account.vue
@@ -5,7 +5,7 @@
             <div class="content fullwidth">
                 <h1>Requested Payments</h1>
 
-                <transition name="fade"><div v-if="errorMessage" v-bind:class="[goodFeedback ? 'goodmessage' : 'message' ]">{{ errorMessage }}</div></transition>
+                <transition name="fade"><div v-if="errorMessage" v-bind:class="[goodFeedback ? 'good.message' : 'message' ]">{{ errorMessage }}</div></transition>
 
                 <div class="accordion">
 
--- a/src/pages/Account.vue
+++ b/src/pages/Account.vue
@@ -5,7 +5,7 @@
             <div class="content fullwidth">
                 <h1>Requested Payments</h1>
 
-                <transition name="fade"><div v-if="errorMessage" v-bind:class="[goodFeedback ? 'good.message' : 'message' ]">{{ errorMessage }}</div></transition>
+                <transition name="fade"><div v-if="message" v-bind:class="[goodFeedback ? 'good message' : 'message' ]">{{ message }}</div></transition>
 
                 <div class="accordion">
 
@@ -69,7 +69,7 @@
         data() {
             return {
                 voucherPayments: Store.trader.pendedVouchers,
-                errorMessage : Store.error,
+                message : Store.error,
                 goodFeedback : false
             }
         },
@@ -79,13 +79,21 @@
         methods: {
             onRequestSubmissionEmail : function(event) {
                 var url = '/traders/' + Store.trader.id + '/voucher-history-email';
-                this.requestEmailBeSent(url,{"submission_date" : event.target.id})
+                this.requestEmailBeSent(url,
+                    {
+                        "submission_date" : event.target.id
+                    }
+                );
             },
             onRequestVoucherHistoryEmail: function() {
                 var url = '/traders/' + Store.trader.id + '/voucher-history-email';
-                this.requestEmailBeSent(url,{ "submission_date" : null});
+                this.requestEmailBeSent(url,
+                    {
+                        "submission_date" : null
+                    }
+                );
             },
-            requestEmailBeSent: function(url,data) {
+            requestEmailBeSent: function(url, data) {
                 // This is a POST, look for the data as a JSON object
                 NetMgr.apiPost(url,data,
                     function (response) {
@@ -101,7 +109,7 @@
                                 mailMsg = "Something went wrong, please try again later.";
                                 console.log(response); // because we need to see what the server said somewhere.
                         }
-                        this.errorMessage = mailMsg;
+                        this.message = mailMsg;
                     }.bind(this),
                     // failure function
                     function (error) {
--- a/src/pages/Account.vue
+++ b/src/pages/Account.vue
@@ -5,7 +5,7 @@
             <div class="content fullwidth">
                 <h1>Requested Payments</h1>
 
-                <transition name="fade"><div v-if="message" v-bind:class="[goodFeedback ? 'good message' : 'message' ]">{{ message }}</div></transition>
+                <transition name="fade"><div v-if="errorMessage" v-bind:class="[goodFeedback ? 'good message' : 'message' ]">{{ errorMessage }}</div></transition>
 
                 <div class="accordion">
 
@@ -69,7 +69,7 @@
         data() {
             return {
                 voucherPayments: Store.trader.pendedVouchers,
-                message : Store.error,
+                errorMessage : Store.error,
                 goodFeedback : false
             }
         },
@@ -109,7 +109,7 @@
                                 mailMsg = "Something went wrong, please try again later.";
                                 console.log(response); // because we need to see what the server said somewhere.
                         }
-                        this.message = mailMsg;
+                        this.errorMessage = mailMsg;
                     }.bind(this),
                     // failure function
                     function (error) {
--- a/src/pages/Scan.vue
+++ b/src/pages/Scan.vue
@@ -112,7 +112,7 @@ export default {
                             response.data.fail.length > 0
                         ) {
                             this.showFail();
-                            this.errorMessage = "[xXx] That voucher may have been used already.";
+                            this.errorMessage = "That voucher may have been used already.";
                         } else  {
                             this.showValidate();
                             this.errorMessage = "";
@@ -121,12 +121,12 @@ export default {
                         Store.getRecVouchers();
                     }.bind(this),
                     // Failure function, hook for error message
-                    () => {
+                    function() {
                         if (!Store.netMgr.online) {
                             this.showQueued();
                             this.queueMessage = "[xXx] Voucher has been added to your queue below.";
                         }
-                    });
+                    }.bind(this));
                 // Do anyway.
                 this.voucherCode = "";
                 this.sponsorCode = "";
--- a/src/pages/Scan.vue
+++ b/src/pages/Scan.vue
@@ -92,7 +92,7 @@ export default {
             if (code.length === parseInt(this.$refs.voucherBox.getAttribute("maxlength"))) {
                 this.$refs.submitVoucher.click();
             }
-        },
+        }
     },
     methods:  {
         onRecordVoucher: function(event) {
--- a/src/pages/Scan.vue
+++ b/src/pages/Scan.vue
@@ -9,7 +9,7 @@
                 <form id="textVoucher" v-on:submit.prevent>
                     <transition name="fade">
                         <div v-if="errorMessage" class="message">{{ errorMessage }}</div>
-                        <div v-if="queueMessage" class="good message">{{ queueMessage }}</div>
+                        <div v-if="queueMessage" class="queue message">{{ queueMessage }}</div>
                     </transition>
                     <label for="sponsorBox" id="lblSponsorBox" class="hidden">Sponsor Code</label>
                     <label for="voucherBox" id="lblVoucherBox" class="hidden">Voucher Code</label>
--- a/src/pages/Scan.vue
+++ b/src/pages/Scan.vue
@@ -163,7 +163,6 @@ export default {
         /**
          * setTimeout is used in these showXYZ methods so the animation is displayed for a meaningful amount of time.
          */
-
         showValidate: function() {
             this.spinner = false;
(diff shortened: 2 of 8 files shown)


ob6160:
Added the offline checking stuff.

src/pages/Tap.vue
```diff
@@ -85,7 +90,12 @@ export default {
                             response.data.fail.length > 0
                         ) {
                             this.showFail();
-                            this.errorMessage = "That voucher may have been used already.";
+                            this.errorMessage = "[xXx] That voucher may have been used already.";
+                        } else if (
+                            !Store.netMgr.online
```
charlesstrange2: ... can't get this to trigger; on ye olde home laptop; check?

charlesstrange2 (dismissed):

src/pages/Account.vue
```diff
@@ -5,7 +5,7 @@
             <div class="content fullwidth">
                 <h1>Requested Payments</h1>
 
-                <transition name="fade"><div v-if="errorMessage" v-bind:class="[goodFeedback ? 'goodmessage' : 'message' ]">{{ errorMessage }}</div></transition>
+                <transition name="fade"><div v-if="errorMessage" v-bind:class="[goodFeedback ? 'good.message' : 'message' ]">{{ errorMessage }}</div></transition>
```
katjam: Is good.message the right syntax to add class="good message"?
Would it be better to default to message and add the good optionally?

Also confused by the semantics - this is an errorMessage - so not good. If it is a good message maybe it should not be passed in a var called errorMessage?

src/pages/Scan.vue
```diff
@@ -7,10 +7,12 @@
...
                 <form id="textVoucher" v-on:submit.prevent>
-                    <transition name="fade"><div v-if="errorMessage" class="message">{{ errorMessage }}</div></transition>
-                    <label for="sponsorBox" id="lblSponsorBox" class="hidden">Sponsor code</label>
-                    <label for="voucherBox" id="lblVoucherBox" class="hidden">Voucher code</label>
-
+                    <transition name="fade">
+                        <div v-if="errorMessage" class="message">{{ errorMessage }}</div>
+                        <div v-if="queueMessage" class="queue message">{{ queueMessage }}</div>
+                    </transition>
+                    <label for="sponsorBox" id="lblSponsorBox" class="hidden">Sponsor Code</label>
```
katjam: I thought I spotted another PR where these were lower cased. 'Sponsor code' - But not sure - just need to check consistency.

src/pages/Scan.vue
```diff
@@ -101,7 +104,12 @@ export default {
                             response.data.fail.length > 0
                         ) {
                             this.showFail();
-                            this.errorMessage = "That voucher may have been used already.";
+                            this.errorMessage = "[xXx] That voucher may have been used already.";
```
katjam: I think we are going to run into some conflicts here with PR #136

src/pages/Scan.vue
```diff
@@ -144,6 +152,15 @@ export default {
             }, 2000);
         },
 
+        showQueued: function() {
+            this.spinner = false;
+            this.queued = true;
+            var self = this;
+            setTimeout(function(){
```
katjam: Don't like arbitrary timeouts. Could we add a comment in here to explain why we need it?

katjam (changes requested):

src/pages/Tap.vue
```diff
@@ -92,16 +99,20 @@ export default {
                             response.data.fail.length > 0
                         ) {
                             this.showFail();
-                            this.errorMessage = "That voucher may have been used already.";
-                        } else {
+                            this.errorMessage = "[xXx] That voucher may have been used already.";
+                        } else  {
```
katjam: Extra space :)

src/pages/Tap.vue
```diff
@@ -92,16 +99,20 @@ export default {
                             response.data.fail.length > 0
                         ) {
                             this.showFail();
-                            this.errorMessage = "That voucher may have been used already.";
-                        } else {
+                            this.errorMessage = "[xXx] That voucher may have been used already.";
```
katjam: This was approved here: https://github.com/neontribe/ARCVMarket/pull/132/files

src/services/netMgr.js
```diff
@@ -203,17 +222,31 @@ NetMgr.setLocalStorageFromToken = function(token) {
 
 // Add interceptor to detect an expired access_token and refresh;
 NetMgr.axiosInstance.interceptors.response.use(
-    function (response) {
+    (response) => {
```
katjam: We are compiling down to ES5 anyway ++ Charlie finds the syntax easier to read. Probably to be consistent we use ES5 here are ES6 in tests? Until we get a standard to enforce - no sense changing arbitrarily.

katjam (changes requested):

src/pages/Scan.vue
ob6160: I've not read this part of the code yet, I'll check to see if it's actually needed.

src/services/netMgr.js
ob6160: :+1:

src/pages/Account.vue
ob6160: I think we should change this so we use a variable called "message" so it could be either way.

katjam: Agree. Go for it. In a new branch or you want to add here?

ob6160: I'll add here

src/pages/Tap.vue
ob6160: What needs changing here? Don't fully understand

katjam: The xXx should not be there.

src/pages/Account.vue
```diff
@@ -69,7 +69,7 @@
         data() {
             return {
                 voucherPayments: Store.trader.pendedVouchers,
-                errorMessage : Store.error,
+                message : Store.error,
```
katjam: Are we happy the Store attr still called error?

src/pages/Scan.vue
```diff
@@ -7,10 +7,15 @@
                 <h1>Scan a voucher code</h1>
 
                 <form id="textVoucher" v-on:submit.prevent>
-                    <transition name="fade"><div v-if="errorMessage" class="message">{{ errorMessage }}</div></transition>
-                    <label for="sponsorBox" id="lblSponsorBox" class="hidden">Sponsor code</label>
-                    <label for="voucherBox" id="lblVoucherBox" class="hidden">Voucher code</label>
-
+                    <transition name="fade">
+                        <div v-if="errorMessage" class="message">{{ errorMessage }}</div>
```
katjam: is this going to be just a message too? and have error class?

src/pages/Scan.vue
```diff
@@ -36,15 +41,15 @@
...
+                        v-bind:class="[{ spinner: this.spinner }, { validate: this.validate }, { fail: this.fail }, { queued: this.queued }]"
                         class="cta"
                     ><span class="hidden offscreen">Submit code</span></button>
 
                 </form>
 
             </div>
 
-            <div v-if="this.vouchers.length > 1">
+            <div v-if="this.vouchers.length >= 1 && !this.netMgr.online">
```
katjam: The queue nees to show only if the voucher list is > 1 I think.

Once online - but not submitted yet, they still need to see it.

src/pages/Scan.vue
```diff
@@ -101,16 +112,20 @@ export default {
                             response.data.fail.length > 0
                         ) {
                             this.showFail();
-                            this.errorMessage = "That voucher may have been used already.";
-                        } else {
+                            this.errorMessage = "[xXx] That voucher may have been used already.";
```
katjam: This copy was approved in a previous PR. Please remove xXx

src/pages/Scan.vue
```diff
@@ -101,16 +112,20 @@ export default {
...
+                        } else  {
                             this.showValidate();
                             this.errorMessage = "";
                         }
                         Store.clearVouchers();
                         Store.getRecVouchers();
                     }.bind(this),
                     // Failure function, hook for error message
-                    function() {
+                    () => {
```
katjam: Another ES6

src/pages/Tap.vue
```diff
@@ -34,15 +39,15 @@
...
+                        v-bind:class="[{ spinner: this.spinner }, { validate: this.validate }, { fail: this.fail }, { queued: this.queued }]"
                         class="cta"
                     ><span class="hidden offscreen">Submit code</span></button>
 
                 </form>
 
            </div>
 
-            <div v-if="this.vouchers.length > 1">
+            <div v-if="this.vouchers.length >= 1 && !this.netMgr.online">
```
katjam: Same as other page.

src/pages/Tap.vue
```diff
@@ -83,16 +91,12 @@ export default {
...
+                        if (response.data.invalid.length > 0) {
                             this.showFail();
                             this.errorMessage = "Please enter a valid voucher code.";
-                        } else if (
-                            response.data.fail.length > 0
-                        ) {
+                        } else if (response.data.fail.length > 0) {
                             this.showFail();
-                            this.errorMessage = "That voucher may have been used already.";
+                            this.errorMessage = "[xXx] That voucher may have been used already.";
```
katjam: Approved

src/pages/Tap.vue
```diff
@@ -101,7 +105,11 @@ export default {
                         Store.getRecVouchers();
                     }.bind(this),
                     // Failure function, hook for error message
-                    function() {
+                    () => {
```
katjam: ES6

src/pages/Tap.vue
```diff
@@ -133,6 +141,15 @@ export default {
             }, 2000);
         },
 
+        showQueued: function() {
+            this.spinner = false;
+            this.queued = true;
+            var self = this;
+            setTimeout(function(){
+                self.queued = false;
+            }, 2000);
```
katjam: use const?

src/sass/app.scss
```diff
@@ -131,22 +131,24 @@ $fa-font-path:"../node_modules/font-awesome/fonts";
...
-.goodmessage {
-  color: $arc_copy;
-  padding-bottom: 1em;
-  &:before {
-    font-weight: bold;
-    font-family: FontAwesome;
-    content: "\f058";
-    color: $arc_success;
-    padding-right: 0.5em;
+  &.error {
```
katjam, line 137: Great but you left a few errors in with just class 'message'

katjam (changes requested):
Lots more little things. Main one is consistent use of error vs errorMessage.

src/pages/Account.vue
ob6160: Yep I checked.

src/pages/Scan.vue
ob6160: This clearly needs more thought - ticket?

src/pages/Scan.vue
ob6160: I don't know, this is Stephs' decision.

src/pages/Tap.vue
ob6160: lol I changed it to const before you commented ;)

src/pages/Scan.vue
```diff
@@ -71,17 +79,20 @@ export default {
...
+            queued: false,
         }
     },
     watch: {
         voucherCode : function(code) {
             if (code.length === parseInt(this.$refs.voucherBox.getAttribute("maxlength"))) {
                 this.$refs.submitVoucher.click();
             }
-        }
+        },
```
katjam: Is that comma intentional?

src/pages/Scan.vue
```diff
@@ -144,22 +160,32 @@ export default {
             this.spinner = true;
         },
 
+        /**
+         * setTimeout is used in these showXYZ methods so the animation is displayed for a meaningful amount of time.
+         */
+
```
katjam: Maybe don't want this extra line

katjam (commented):
Pretty well there.

src/pages/Scan.vue
ob6160: Trailing commas are allowed in JS objects since ES5, don't see why we shouldn't have it :) It's just JSON which doesn't like them

katjam: Trailing commas are good for a few reasons - but inconsistent use of them negates the goodness. So I was just checking because it looked inconsistent.

katjam (approved):
If the tests pass :pushpin:

src/pages/Scan.vue
ob6160: ah, the goodness is being negated then, I'll remove it,

katjam:
ha ha.

The goodness I mean for me is:

1- don't get a false change in commit (line appears to have changed because
of adding comma which make real change less clear
2- can't forget to add comma between previous last item and new last item
in a multi-line list - which is an easy syntax XX to make

ob6160:
yep, I've gotten into the habit of doing it as it's a Drupal standard too

PR #133, opened 2017-06-22
Add validation and make slashes change focus from sponsor to voucher box

gingerbreadpup:
Seems to work as it should now - let me know if I've missed something...


katjam (approved):
Looks good to me. Shame we keep repeating so much on the tap and scan pages though. Needs a refactor next time.
:fish:

gingerbreadpup:
Yep - I'll check that there's a card for it about refactoring the code.

PR #134, opened 2017-06-22
Initial offline checking draft

ob6160:


src/services/netMgr.js
```diff
@@ -203,17 +222,31 @@ NetMgr.setLocalStorageFromToken = function(token) {
...
         // everything fine! return the response to apiGet/apiPost
         return response;
     },
-    function (error) {
+    (error) => {
         // Get the original request configuration.
         var origCfg = error.config;
 
         // Get the error from that request.
         var origResp = error.response;
```
katjam, line 237: Is this the status?

src/services/netMgr.js
```diff
@@ -203,17 +222,31 @@ NetMgr.setLocalStorageFromToken = function(token) {
...
 
+        // Grab the error message.
+        var errMsg = error.message || null;
```
katjam, line 240: and this the message?

src/services/netMgr.js
```diff
@@ -203,17 +222,31 @@ NetMgr.setLocalStorageFromToken = function(token) {
...
+
+        // Default to assuming we are online.
+        NetMgr.setOnlineStatus(true);
+
+        // Set offline if a Network Error occurs. There should be a more specific error message but I couldn't find one.
+        if(errMsg === 'Network Error') {
```
katjam, line 246: Would it be better to use a const than a string match?

katjam (approved):
Yup - willing to give this a go if you think it works. We have the Beta trial to find the bugs.

charlesstrange2 (approved):
give it a whirl!
:shipit:

src/services/netMgr.js
ob6160: This has a more verbose version of error.message so I used error.message instead. I want to find a better way of checking of we're offline than just "Network Error" but this will do fine for now.

src/services/netMgr.js
ob6160: Again, something which needs looking into. I think there's a better thing to compare against.

PR #136, opened 2017-06-23
1.0/119 add local storage capability traders

charlesstrange2:
Doh! conflicts!

This refactors store.recVouchers and store.vouchers into store.trader, because we're keen to stash those two values in localstorage for offline mode, and previous work has already sorted stashing Trader on various events.

### DIFF
--- a/src/pages/Scan.vue
+++ b/src/pages/Scan.vue
@@ -56,11 +56,12 @@ export default {
         Profile
     },
     data: function() {
+        console.log(Store);
         return {
             sponsorCode : "",
             voucherCode : "",
-            vouchers : Store.vouchers,
-            recVouchers : Store.recVouchers,
+            vouchers : Store.trader.vouchers,
+            recVouchers : Store.trader.recVouchers,
             errorMessage : Store.error,
             spinner: false,
             validate: false,
--- a/src/pages/Scan.vue
+++ b/src/pages/Scan.vue
@@ -84,20 +84,32 @@ export default {
                     // Success function
                     function(response) {
                         // Add error message for invalid and fail codes.
-                        if (
-                            response.data.invalid.length > 0
-                        ) {
+                        var data = response.data;
+
+                        if (data.invalid.length + data.fail.length == 1) {
+                            // single mismatch handler;
+                            if (data.invalid.length > 0) {
+                                this.showFail();
+                                this.errorMessage = "[xXx] Please enter a valid voucher code.";
+
+                            } else if (data.fail.length > 0) {
+                                this.showFail();
+                                this.errorMessage = "[xXx] That voucher may have been used already.";
+                            }
+
+                        } else if (data.invalid.length + data.fail.length > 1) {
+                            // rough multifailure manager
                             this.showFail();
-                            this.errorMessage = "[xXx] Please enter a valid voucher code.";
-                        } else if (
-                            response.data.fail.length > 0
-                        ) {
-                            this.showFail();
-                            this.errorMessage = "[xXx] That voucher may have been used already.";
+                            this.errorMessage = "[xXx] "+
+                                data.success.length+" accepted, " +
+                                data.fail.length + " rejected and " +
+                                data.invalid.length +" were invalid.";
                         } else {
+                            // all in!
                             this.showValidate();
                             this.errorMessage = "";
                         }
+                        // The server has processed our list, clear it.
                         Store.clearVouchers();
                         Store.getRecVouchers();
                     }.bind(this),
--- a/src/pages/Scan.vue
+++ b/src/pages/Scan.vue
@@ -68,8 +68,8 @@ export default {
         return {
             sponsorCode : "",
             voucherCode : "",
-            vouchers : Store.vouchers,
-            recVouchers : Store.recVouchers,
+            vouchers : Store.trader.vouchers,
+            recVouchers : Store.trader.recVouchers,
             errorMessage : Store.error,
             spinner: false,
             validate: false,
@@ -108,9 +108,13 @@ export default {
                         } else if (data.invalid.length + data.fail.length > 1) {
                             // rough multifailure manager
                             this.showFail();
-                            this.errorMessage = `[xXx] ${data.success.length} accepted,
-                                ${data.fail.length} rejected
-                                and ${data.invalid.length} were invalid.`;
+                            this.errorMessage = "[xXx] " +
+                                data.success.length +
+                                " accepted, " +
+                                data.fail.length +
+                                " rejected and " +
+                                data.invalid.length +
+                                " were invalid.";
                         } else {
                             // all in!
                             this.showValidate();
--- a/src/pages/Scan.vue
+++ b/src/pages/Scan.vue
@@ -86,7 +86,7 @@ export default {
                         // Add error message for invalid and fail codes.
                         var data = response.data;
 
-                        if (data.invalid.length + data.fail.length == 1) {
+                        if (data.invalid.length + data.fail.length === 1) {
                             // single mismatch handler;
                             if (data.invalid.length > 0) {
                                 this.showFail();
@@ -114,7 +114,9 @@ export default {
                         Store.getRecVouchers();
                     }.bind(this),
                     // Failure function, hook for error message
-                    function() {
+                    function(error) {
+                        //network error of some kind;
+                        //don't clear the voucherlist!
                     });
                 // Do anyway.
                 this.voucherCode = "";
--- a/src/pages/Scan.vue
+++ b/src/pages/Scan.vue
@@ -108,13 +108,14 @@ export default {
                         } else if (data.invalid.length + data.fail.length > 1) {
                             // rough multifailure manager
                             this.showFail();
-                            this.errorMessage = "[xXx] " +
-                                data.success.length +
-                                " accepted, " +
-                                data.fail.length +
-                                " rejected and " +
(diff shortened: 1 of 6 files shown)


charlesstrange2:
Updated with release branch ;-/

src/store.js
```diff
@@ -153,7 +157,7 @@ store.getRecVouchers = function () {
             var newVouchers = Object.keys(response.data).map(function(k){
                 return response.data[k];
             });
-            this.mergeRecVouchers(newVouchers);
+            this.mergeRecVouchers.call(this, newVouchers);
```
ob6160: We can change this back since it was just for when we were debugging.

src/store.js
```diff
@@ -164,23 +168,27 @@ store.getRecVouchers = function () {
...
 
 /**
  * Adds a voucher code and submits it.
  */
 store.addVoucherCode = function (voucherCode, success, failure) {
-    this.vouchers.push(voucherCode);
-    this.transitionVouchers('collect', this.vouchers, success, failure);
+    this.trader.vouchers.push(voucherCode);
+    // store the whole trader
+    localStorage['Store.trader'] = JSON.stringify(this.trader);
```
ob6160: I'd rather we keep all localStorage sets for store in one place. Perhaps we could have a method similar to "setLocalStorageFromToken" in NetMgr in Store.

src/pages/Scan.vue
```diff
@@ -92,25 +92,39 @@ export default {
...
+                                this.showFail();
+                                this.errorMessage = "That voucher may have been used already.";
+                            }
+
+                        } else if (data.invalid.length + data.fail.length > 1) {
+                            // rough multifailure manager
                             this.showFail();
-                            this.errorMessage = "That voucher may have been used already.";
+                            this.errorMessage = "[xXx] "+
+                                data.success.length+" accepted, " +
```
katjam: Formatting

katjam: Do we not want to tell them the codes for each?

src/pages/Scan.vue
```diff
@@ -92,25 +92,39 @@ export default {
...
                         Store.clearVouchers();
                         Store.getRecVouchers();
                     }.bind(this),
                     // Failure function, hook for error message
-                    function() {
+                    function(error) {
+                        //network error of some kind;
+                        //don't clear the voucherlist!
                     });
                 // Do anyway.
```
katjam: Indenting?

src/pages/Tap.vue
```diff
@@ -83,25 +83,39 @@ export default {
...
-                            response.data.invalid.length > 0
-                        ) {
-                            this.showFail();
-                            this.errorMessage = "Please enter a valid voucher code.";
-                        } else if (
-                            response.data.fail.length > 0
-                        ) {
+                        var data = response.data;
+
+                        if (data.invalid.length + data.fail.length == 1) {
```
katjam, line 88: This needs a refactor. Scan Tap and Upload will potentially use this same system.

src/store.js
```diff
@@ -164,23 +176,27 @@ store.getRecVouchers = function () {
  */
 store.mergeRecVouchers = function (replacements) {
     // this zeros the array and re-add things in a vue-friendly way
-    this.recVouchers.splice(0, this.recVouchers.length, replacements);
+    this.trader.recVouchers.splice(0,this.trader.recVouchers.length, replacements);
+    // changed the recVouchers! Quick, save them!
```
katjam: Capital C?

src/store.js
```diff
@@ -164,23 +176,27 @@ store.getRecVouchers = function () {
  */
 store.mergeRecVouchers = function (replacements) {
     // this zeros the array and re-add things in a vue-friendly way
-    this.recVouchers.splice(0, this.recVouchers.length, replacements);
+    this.trader.recVouchers.splice(0,this.trader.recVouchers.length, replacements);
```
katjam: Space after the 0

src/store.js
```diff
@@ -192,17 +208,16 @@ store.pendRecVouchers = function (success,failure) {
  */
 store.clearVouchers = function () {
     // alter current array, not swap for new one or vue gets sad!
```
katjam: Capital A?

src/store.js
```diff
@@ -192,17 +208,16 @@ store.pendRecVouchers = function (success,failure) {
...
-    this.vouchers.splice(0, this.vouchers.length);
+    this.trader.vouchers.splice(0, this.trader.vouchers.length);
+    // trader has changed, alter it.
```
katjam: Capital T?

katjam (changes requested):

src/pages/Scan.vue
```diff
@@ -92,25 +92,39 @@ export default {
...
+                            }
+
+                        } else if (data.invalid.length + data.fail.length > 1) {
+                            // rough multifailure manager
                             this.showFail();
-                            this.errorMessage = "That voucher may have been used already.";
+                            this.errorMessage = "[xXx] "+
+                                data.success.length+" accepted, " +
+                                data.fail.length + " rejected and " +
+                                data.invalid.length +" were invalid.";
```
katjam: missing space after +

src/pages/Scan.vue
```diff
@@ -92,25 +92,39 @@ export default {
...
+                            } else if (data.fail.length > 0) {
+                                this.showFail();
+                                this.errorMessage = "That voucher may have been used already.";
+                            }
+
+                        } else if (data.invalid.length + data.fail.length > 1) {
+                            // rough multifailure manager
                             this.showFail();
-                            this.errorMessage = "That voucher may have been used already.";
+                            this.errorMessage = "[xXx] "+
```
katjam: These + should be leading rather than trailing - unless js people like it this way?

src/pages/Scan.vue
```diff
@@ -92,25 +92,38 @@ export default {
...
+                            } else if (data.fail.length > 0) {
+                                this.showFail();
+                                this.errorMessage = "That voucher may have been used already.";
+                            }
+
+                        } else if (data.invalid.length + data.fail.length > 1) {
+                            // rough multifailure manager
                             this.showFail();
-                            this.errorMessage = "That voucher may have been used already.";
+                            this.errorMessage = `[xXx] ${data.success.length} accepted,
```
katjam: Are you sure there is support for this? We are supporting older browsers. Rather not need a polyfill if we can help it. Maybe it compiles fine... but I see no good reason to introduce.

I didn't mind the concatenation... it was just badly formatted.

katjam: Ultimately - we will factor this out anyway - once we add the codes into it for https://trello.com/c/0qJxIWwn/358-better-error-messages-for-multiple-voucher-submit

katjam (changes requested):

src/pages/Scan.vue
ob6160: Isn't that what babel is there for - so we can use these helpful new language features? In my mind it's fine because we're refactoring this anyway.

katjam: I agree - but there is some evidence that babel alone doesn't fix this entirely - and I figured not worth the risk.

ob6160: ok - I'll change it back plus formatting :fork_and_knife:

katjam: sorry :(

src/pages/Scan.vue
```diff
@@ -92,25 +92,42 @@ export default {
...
+                            } else if (data.fail.length > 0) {
+                                this.showFail();
+                                this.errorMessage = "That voucher may have been used already.";
+                            }
+
+                        } else if (data.invalid.length + data.fail.length > 1) {
+                            // rough multifailure manager
                             this.showFail();
-                            this.errorMessage = "That voucher may have been used already.";
+                            this.errorMessage = "[xXx] " +
```
katjam: is it standard to add the concatinator at the end in js?

src/pages/Scan.vue
```diff
@@ -92,25 +92,42 @@ export default {
...
+                                data.success.length +
+                                " accepted, " +
+                                data.fail.length +
+                                " rejected and " +
+                                data.invalid.length +
+                                " were invalid.";
```
katjam: I'd move the ; down to the next line :)

katjam (changes requested):
so close.

katjam:
@ob6160 I think this is good to go as long as you like the + trailing.

src/pages/Scan.vue
ob6160: :nerd_face: I'm intrigued, I've never seen the "';" on a separate line apart from in the testcafe tests. If anything I find it harder to read when it's on a separate line and I've never seen it done in any other JS project I've worked on.

src/pages/Scan.vue
ob6160: It feels like "the right thing to do"™... It's what I'd do in any other language.

katjam: Most languages is like:

```
$txt = "blah"
    + "blah"
    + "blah"
;
```

src/pages/Scan.vue
katjam: It's so you can easily see where the start and end of a multi-line is more easily.

It also helps with git diffs when stuff is added to the end so it doesn't appear that 2 lines have changed.

Same theory as }); on a new line

katjam (approved):
:do_not_litter:

PR #138, opened 2017-06-23
1.0/commit styling fix

alexalala:
https://trello.com/c/CLlrPHLZ/355-styling-for-commit-hash-info-rethink

footer wouldn't position itself below the main content without being within main, which I didn't want so I created this lil workaround which works a lot better I think.

🐌


katjam (approved):
looking pretty. :sun_behind_rain_cloud:

PR #139, opened 2017-06-25
delete added and voucher number increased

alexalala:
Had to restart initially as it couldn't connect to run npm test (like the browser problem we had) but then passed first time which is a good sign!


katjam (approved):
Seems crazy overkill for unpredictable test runs... but I know we agreed if it works do it.

Could you stick a comment above one of those lines that says why we are doing it? (Seems unpredictable whether test runner clicks at start or end of text)

alexalala:
@katjam yep will do!

PR #140, opened 2017-06-26
fixed a previously fixed bug that doesnt seem to work anymore... now …

alexalala:
…it does yay!

https://trello.com/c/9bdMXsD1/193-bug-submit-voucher-all-browsers-doesnt-activate-the-submit-button

🐭

### DIFF
--- a/src/pages/Scan.vue
+++ b/src/pages/Scan.vue
@@ -249,6 +249,10 @@ export default {
                 }
                 return;
             }
+            //allow enter key to submit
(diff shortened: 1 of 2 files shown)


ob6160 (approved):
Looks good :ice_cream:

katjam:
Do you need a similar change on scan? Or does that form auto submit when the scan is complete?

katjam (commented):
If it's a problem for scan also, please change there too - otherwise. Go for it.

PR #141, opened 2017-06-26
1.0/362 offline interface

gingerbreadpup:
Passing over to Ollie


katjam (approved):
Yup. That looks better. If it works. :shipit: Along with some clear acceptance on the card so we know what intended behaviour to test.

src/components/Queue.vue
```diff
@@ -90,12 +91,7 @@ export default {
     },
     computed: {
         currentlyShown: function() {
-            return (
-                this.fail
-                || this.validate
-                || (this.vouchers.length >= 1 && !this.netMgr.online)
-                || this.queue.sendingStatus
-            );
+            return (this.fail || this.validate || (this.vouchers.length >= 1 && !this.netMgr.online) || this.queue.sendingStatus);
```
katjam: I think split this back to multiline so we can see more clearly what is going on.

src/components/Queue.vue
```diff
@@ -37,20 +46,95 @@
...
+                this.validate = false;
+                this.message = '';
+            }.bind(this), RESULT_TIMER);
+        },
+
+        showFail: function() {
+            this.spinner = false;
+            this.fail = true;
+
+            this.message = "[xXx] Whoops! There may be a network problem. When you have a better signal, click 'Submit queued vouchers' to retry.";
```
katjam, line 120: @gingerbreadpup At this stage let's get the text in as approved if we can so we don't have to card another xXx sweep.

katjam (changes requested):
Just a couple of small changes - but let's get this in asap so we can test manually.

Do we have any browser tests for it - or too hard to mock the offline status?

src/components/Queue.vue
```diff
@@ -106,6 +111,8 @@ export default {
             this.message = "[xXx] Thanks! We've successfully submitted your queued vouchers.";
```
katjam: Oh another xXx - maybe we do need a sweep again?

katjam (approved):
looks promising! :gem:

src/components/Queue.vue
gingerbreadpup: I'm happy with this text.

src/components/Queue.vue
```diff
@@ -1,15 +1,24 @@
...
+    <div class="content narrow queuedVouchers">
 
         <h1 v-on:click="collapsed = !collapsed" class="expandable queue" v-bind:class="{'expanded' : !collapsed}">Queued vouchers</h1>
 
         <transition name="fade" v-if="show">
-            <div class="goodmessage queue">
-                [xXx] Thanks! We can't confirm the <strong>{{ vouchers.length }}</strong> vouchers just now, there may be a network problem. Keep adding to this list until they all go through.
+            <div v-if="!message" class="goodmessage queue">
+                Thanks! We can't confirm the <strong>{{ vouchers.length }}</strong> vouchers just now, there may be a network problem. When you have a better signal, click 'Submit queued vouchers' to try and send them again.
+            </div>
```
gingerbreadpup: I think this message should just be "You have x vouchers in your queue". The current message above would be confusing to see if you were viewing your queue containing vouchers that had been added to it during an earlier session. If you want I can do a PR for this after you merge in.

I also think the 'Voucher added to your queue' message should be revised to indicate there being a lack of signal - I'll add this to my PR too (and will make a card for it).

PR #144, opened 2017-06-26
footer repositioned when error appears

alexalala:
https://trello.com/c/VOj0gp9y/364-bug-footer-clips-login-button-and-main-content-on-small-screen-sizes-iphone-5-and-6

🐩


katjam (commented):
Seems a bit round-a-bout... I'd prefer a css fix rather than adding a class in this way - but if @steph thinks it is good to go this way because there is no simple answer. OK by me.

katjam (approved):
OK :shamrock:

gingerbreadpup (changes requested):
You can do this without the js - I made a PR here just so you can see the changes I made in one place - you can close that one whenever.

https://github.com/neontribe/ARCVMarket/pull/147

katjam (changes requested):
Let's ditch this one?

PR #147, opened 2017-06-26
1.0/footer tweak

gingerbreadpup:
Just made this PR so you can see another way of doing it without javascript - it was mainly the container being position:absolute that was making the footer not respect the container's size so once I changed that to relative it all worked a bit better. You can close this whenever!


katjam (approved):
@gingerbreadpup As long as the version number doesn't end up in the password field - I say we keep this one over @alexalala's one.

alexalala:
@gingerbreadpup @katjam Yep just tested and works beautifully! I'll close my PR if you wanna fix Travis and merge this one?

PR #148, opened 2017-06-27
img path changed

alexalala:
as when it's built into dist it's available on the root

https://trello.com/c/5MWXvyMZ/310-style-noscript-tags-for-when-javascript-is-disabled

🐐


katjam (approved):
:camera:

PR #150, opened 2017-06-27
1.0/375 clear queuemessage

katjam:
I think we should probably also clear the message when the queue is successfully submitted - but up to @gingerbreadpup because this is a different component. For now at least this solves the issue of other messages and errors never triggering after a queueMessage was instantiated.

There is another ticket for next sprint to refactor our messaging system - so that we just have messages rather than messages and errors which might help too.


gingerbreadpup (approved):
Looks good to me! :shipit: :dancer:

PR #151, opened 2017-06-27
1.0/fix footer

gingerbreadpup:
Ok... I could not make Travis happy. There was so much stuff breaking just because I used position: relative in my last PR, which is just insane and makes no sense to me at all. This is the only css-only way I can seem to do it but I think it's fine.

### DIFF
--- a/src/pages/Login.vue
+++ b/src/pages/Login.vue
@@ -21,12 +21,13 @@
 
             </div>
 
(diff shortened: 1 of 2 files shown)


alexalala (commented):
looks amazing :shipit:

alexalala (approved):
looks amazing :shipit:

PR #152, opened 2017-06-28
1.0/377 success bug

gingerbreadpup:
This PR is for [this card](https://trello.com/c/d3k6wtpp/377-bug-the-success-message-to-user-on-submit-voucher-queue-is-misleading-we-need-to-clear-previous-not-enough-signal-error-too)

Probably isn't the most elegant code, but the queue is now functioning as it should - as far as I can tell!

* The 'Not enough signal, voucher queued' message now clears when you click the the 'Submit queued vouchers' button.
* On successful queue submission, a more detailed error message displays, letting the user know how many vouchers were accepted, duplicates or invalid.
* Invalid vouchers weren't being counted properly (they were just undefined), these are now counted and displayed in the feedback message
* Moved queue feedback code out of tap and scan, and into the queue component
* Tweaked styling for the voucher code list

### DIFF
--- a/src/components/Queue.vue
+++ b/src/components/Queue.vue
@@ -137,6 +137,7 @@ export default {
                 var fail = '';
                 var invalid = '';
 
+                // Construct the feedback message.
                 if (data.success.length === 1) {
                     success = "1 voucher was accepted, ";
                 } else {
@@ -163,6 +164,7 @@ export default {
                 ;
 
                 this.showValidate();
+                // Update the parent component so that it can hide the not enough signal message
                 this.$emit('update', this.clearMessage);
 
             }.bind(this),
--- a/src/components/Queue.vue
+++ b/src/components/Queue.vue
@@ -164,7 +164,7 @@ export default {
                 ;
 
                 this.showValidate();
-                // Update the parent component so that it can hide the not enough signal message
+                // Send out an update which will be picked up in tap, so that it can hide the not enough signal message
                 this.$emit('update', this.clearMessage);
 
             }.bind(this),
--- a/src/components/Queue.vue
+++ b/src/components/Queue.vue
@@ -155,7 +155,6 @@ export default {
                     invalid = "and " + data.invalid.length + " were invalid.";
                 }
 
-                this.showValidate();
                 this.message
                     = "Thanks! Your queue has been successfully submitted. "
                     + success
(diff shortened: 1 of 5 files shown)


src/store.js
```diff
@@ -201,11 +201,11 @@ store.mergeRecVouchers = function (replacements) {
 /**
  * Adds a voucher code and submits it.
  */
-store.addVoucherCode = function (voucherCode, success, failure) {
+store.addVoucherCode = function (voucherCode, success, failure, invalid) {
     this.trader.vouchers.push(voucherCode);
     // Store the whole trader
     this.setLocalStorageFromUserTraders();
-    this.transitionVouchers('collect', this.trader.vouchers, success, failure);
+    this.transitionVouchers('collect', this.trader.vouchers, success, failure, invalid);
```
katjam: I don't think the invalid is needed here or line 204... The success and failure being passed thorough should refer to whether or not the request gets a successful response as opposed to the quality of the respoonse (whether it contains good and bad code messages)

katjam (commented):
Happy to merge this in - as it looks like it works but we should flag up for a re-visit. As @gingerbreadpup points out it is a bit over complicated.

src/pages/Tap.vue
```diff
@@ -248,6 +242,11 @@ export default {
             return String.fromCharCode(charCode);
         }
     },
+    computed: {
+        showQueueMsg() {
```
ob6160, line 246: Is a computed value necessary here?

src/pages/Scan.vue
```diff
@@ -260,6 +253,11 @@ export default {
             return String.fromCharCode(charCode);
         }
     },
+    computed: {
```
ob6160, line 257: Same as in Tap

src/components/Queue.vue
```diff
@@ -127,12 +127,46 @@ export default {
...
             this.startSpinner();
 
-            Store.transitionVouchers('collect', this.vouchers, function() {
+            Store.transitionVouchers('collect', this.vouchers, function(response) {
                 // The server has processed our list, clear it.
                 Store.clearVouchers();
                 Store.getRecVouchers();
 
+                var data = response.data;
+                var success = '';
```
ob6160, line 139: Just a style thing, it might read better if these were initialised like this as they're all being set to the same value.
var success = '', fail = '', invalid = '';

katjam (approved):

ob6160:
whoops I never submitted my review :nerd_face:

katjam:
lol. Yep. We are due a big refactor anyway. Don't worry.

PR #153, opened 2017-06-29
1.0/red errors

alexalala:
https://trello.com/c/77e3go6m/378-error-messages-should-be-in-red-information-messages-should-stay-in-black

Just checked and it shouldn't mess up @gingerbreadpup 's PR as the message that returns the successful/failed vouchers will never (for now anyway) be an error message.

🐰

### DIFF
--- a/src/components/Queue.vue
+++ b/src/components/Queue.vue
@@ -5,11 +5,14 @@
         <h1 v-on:click="collapsed = !collapsed" class="expandable queue" v-bind:class="{'expanded' : !collapsed}">Queued vouchers</h1>
 
         <transition name="fade" v-if="show">
-            <div v-if="!message" class="goodmessage queue">
-                You have <strong>{{ vouchers.length }}</strong> vouchers in your queue.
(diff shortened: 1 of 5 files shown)


src/components/Queue.vue
```diff
@@ -5,12 +5,15 @@
         <h1 v-on:click="collapsed = !collapsed" class="expandable queue" v-bind:class="{'expanded' : !collapsed}">Queued vouchers</h1>
 
         <transition name="fade" v-if="show">
-            <div v-if="!message" class="goodmessage queue">
-                You have <strong>{{ vouchers.length }}</strong> vouchers in your queue.
+            <div v-if="fail && message" class="goodmessage error">
```
katjam: Are you sure? goodmessage and error?

@gingerbreadpup For start of next sprint - would be good to define what types of messages we can have and how to implement them. We seem to be going in circles on this every time a new person implements a message.

katjam (changes requested):

katjam (approved):
:seedling: Awesome!

PR #154, opened 2017-06-29
Prevent the enter key sent from the scanner from causing the form to …

ob6160:
…be submitted twice

I also changed the watch event so we call the"onRecordVoucher" method directly for the sake of clarity and consistency.


katjam (approved):
Seems OK to me. :shamrock:

PR #155, opened 2017-06-29
Fix queue message bug

gingerbreadpup:
The queue voucher count and success/fail messages were hiding behind the queue submit button, this fixes that so they display above it.

[Card 1](https://trello.com/c/ZhXFInTW/398-bug-no-message-to-say-how-many-vouchers-are-in-your-queue)
[Card 2](https://trello.com/c/y7ntV2gg/399-bug-queue-feedback-messages-displaying-behind-queue-submit-button)

### DIFF
--- a/src/sass/app.scss
+++ b/src/sass/app.scss
@@ -456,14 +456,6 @@ main > *:target ~ * {
 
 #textVoucher,
 .queuedVouchers {
-  label {
-    margin-bottom: 0;
-    height: 34px;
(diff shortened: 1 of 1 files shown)


katjam:
@gingerbreadpup Good catch - Karl and I saw that as we were leaving the office this evening.

katjam (approved):
Going to assume it does what you say!
:leopard:

PR #160, opened 2017-07-03
Move vouchers into objects and store online state for queue display p…

ob6160:
…urposes

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -39,6 +39,28 @@ store.resetStore = function() {
 };
 
 /**
+ * Returns a flat array of voucher codes.
+ *
+ * @param trader
+ *   Currently not used. Will be needed when we support the storage of multiple traders.
+ * @returns {Array}
+ */
+store.getTraderVoucherList = function(trader) {
+    return this.trader.vouchers.map((v) => v.code);
+};
+
+/**
+ * Indicates whether any vouchers have been added whilst the user was offline.
+ *
+ * Returns false if any of the vouchers were recorded whilst the user was offline.
+ *
+ * @returns {Boolean}
+ */
+store.getVouchersOnlineStatus = function() {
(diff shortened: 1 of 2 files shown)


charlesstrange2 (approved):
Looks good
:shipit:

src/store.js
```diff
@@ -39,6 +39,28 @@ store.resetStore = function() {
...
+
+/**
+ * Indicates whether any vouchers have been added whilst the user was offline.
+ *
+ * Returns false if any of the vouchers were recorded whilst the user was offline.
+ *
+ * @returns {Boolean}
+ */
+store.getVouchersOnlineStatus = function() {
+    return !!(this.trader.vouchers.reduce((v, e) => e.online, true));
```
katjam: In what case would `this.trader.vouchers.reduce((v, e) => e.online, true)` not return the same value as the !! of it?

Will the reduce always return true or false or can it be undefined or a string? Why do you need to feed it an initial value of true?

ob6160: At some point e.online has the potential to be "undefined". I used "!!" in order to work around this.

katjam: If the aim is to detect whether the vouchers array contains a voucher with online: false then why not just look for that?

At what point is e.online undefined? I think it is important not to mask things. Especially in js because that's how crazy bugs happen.

katjam: And the initial true? When does this `this.trader.vouchers.reduce((v, e) => e.online, true)` differ from `this.trader.vouchers.reduce((v, e) => e.online)`?

katjam: Is the function in some way comparing the value to true? `this.trader.vouchers.reduce((v, e) => e.online === true)` I am confused by the reduce. Obviously you don't want just the last value of the array's online attribute. So how does the reduce chain the booleans?

katjam (approved):
Pending a second pair of eyes over getVouchersOnlineStatus

PR #162, opened 2017-07-07
Add titles to input icons in the toolbar

gingerbreadpup:
[Card](https://trello.com/c/WvPZyKpg/382-keying-in-icons-should-there-be-text-displaying-when-user-hovers-over-icons)


charlesstrange2 (approved):
Great
:icecream:

PR #165, opened 2017-07-11
Submit scanned voucher after one second and fade out inputted text ov…

ob6160:
…er this time period

### DIFF
--- a/src/pages/Scan.vue
+++ b/src/pages/Scan.vue
@@ -23,8 +23,10 @@
                             minlength="2"
                             maxlength="5"
                             autofocus="autofocus"
+                            v-bind:class="{ 'input-text-hidden': queued }"
                         >
                         <input id="voucherBox"
+                            @keydown.enter.prevent
                             v-on:keyup.delete='onDelVoucherBox'
                             @keypress='onKeypressVoucherBox'
                             type="tel"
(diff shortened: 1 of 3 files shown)


src/pages/Scan.vue
```diff
@@ -62,7 +65,7 @@ import Store from '../store.js';
 import Profile from '../components/Profile.vue';
 import Queue from '../components/Queue.vue';
 
-const RESULT_TIMER = 2000;
+const RESULT_TIMER = 1000;
```
charlesstrange2: total nitpick: SCAN_TIMEOUT, or similar?

charlesstrange2: We're avoiding ES6 at the moment :cry: just in case :fear:

charlesstrange2 (dismissed):
A trivial nitpick or two and we need a test script.

src/pages/Scan.vue
ob6160: I think we should keep this the same name as the const in Tap.vue as we want to - at some point - share functionality between the two. We'll probably have to change it back then anyway.

charlesstrange2:
Good point - there's a long grass card about combining tap and scan in one
thing.

src/pages/Scan.vue
ob6160: Changed

gingerbreadpup (commented):
Yep :shipit:

gingerbreadpup (approved):

PR #167, opened 2017-07-12
Add additional copy in the voucher queue

gingerbreadpup:
This might get changed later - but just putting something in for now with the xXx.

### DIFF
--- a/src/components/Queue.vue
+++ b/src/components/Queue.vue
@@ -12,7 +12,7 @@
                 {{ message }}
             </div>
             <div v-else class="message">
-                You have <strong>{{ vouchers.length }}</strong> vouchers in your queue.
+                [xXx]You have <strong>{{ vouchers.length }}</strong> vouchers in your queue. These voucher codes will be checked when you submit your queue.
             </div>
         </transition>
 
--- a/src/components/Queue.vue
+++ b/src/components/Queue.vue
@@ -12,7 +12,7 @@
                 {{ message }}
             </div>
             <div v-else class="message">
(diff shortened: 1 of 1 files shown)


katjam (approved):
Looks good to me. :+1:

src/components/Queue.vue
```diff
@@ -12,7 +12,7 @@
                 {{ message }}
             </div>
             <div v-else class="message">
-                You have <strong>{{ vouchers.length }}</strong> vouchers in your queue.
+                [xXx]You have <strong>{{ vouchers.length }}</strong> vouchers in your queue. These voucher codes will be checked when you submit your queue.
```
charlesstrange2: Do we want to be cunning with "1 vouchers"?

katjam: There may be built in pluralisation library in vue. I know other frameworks do provide by default.

charlesstrange2: vuejs 2.0 removed filters like pluralize while on some "purity" purge. The docs (https://vuejs.org/v2/guide/migration.html) think https://www.npmjs.com/package/pluralize is a good replacement.

gingerbreadpup: I think for this card, either I should add a line of code in with a condition in to add the s - or make a separate card about pluralisation because there are a few other places in the app that would need to be fixed at the same time if we're going to use a package. What do you think?

charlesstrange2: agreed - by hand here and a fresh card (that might get put under the ribbon) for the rest.

charlesstrange2 (approved):
Add tests, then 
:shipit:

PR #168, opened 2017-07-12
1.1/144 reset functionality

gingerbreadpup:
So nothing is wired up yet, but I've created pages for requesting a password, and then changing it the password.

Up to you if you want to approve or just close and pick up the branch to add the back-end bits! :+1:

### DIFF
--- a/src/components/Masthead.vue
+++ b/src/components/Masthead.vue
@@ -3,8 +3,8 @@
     <div class="wrapper">
 
         <header>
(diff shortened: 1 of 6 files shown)


katjam:
Yep - this looks ok. I think you can put it in and we can always then tweak if it doesn't quite gel with the backend - once that is done.

katjam:
Oops totally did not mean to do that!

katjam (approved):
Yep - merge this is an put a priority on getting the backend wired up.

PR #169, opened 2017-07-12
Move all messages into a new Message component.

ob6160:
Also taking the opportunity to try Vue mixins which will allow us to greatly reduce code duplication.

### DIFF
--- a/src/components/Message.vue
+++ b/src/components/Message.vue
@@ -1,0 +1,35 @@
+<template id="message">
+  <transition name="fade" v-if="visible">
+    <div class="message" v-bind:class="{'error': error, 'good': success}" v-html="text"></div>
+  </transition>
+</template>
+
+<script>
+  import constants from '../constants';
+
+  export default {
+      name: 'message',
+      props: {
+          text:  { default: '' },
+          state: { default: constants.MESSAGE_STATUS },
+          shown: { default: true },
+      },
+      data: function() {
+          return {}
+      },
+      computed: {
+          error: function() {
+              return this.state === constants.MESSAGE_ERROR;
+          },
+          status: function() {
+              return this.state === constants.MESSAGE_STATUS;
+          },
+          success: function() {
+              return this.state === constants.MESSAGE_SUCCESS;
+          },
+          visible: function() {
+              return (this.shown && this.text);
+          }
+      }
+  }
+</script>
--- a/src/components/Message.vue
+++ b/src/components/Message.vue
@@ -1,6 +1,6 @@
 <template id="message">
   <transition name="fade" v-if="visible">
-    <div class="message" v-bind:class="{'error': error, 'good': success}" v-html="text"></div>
+    <div class="message" v-bind:class="{'error': error, 'success': success}" v-html="text"></div>
   </transition>
 </template>
 
--- a/src/components/Queue.vue
+++ b/src/components/Queue.vue
@@ -4,7 +4,7 @@
 
         <h1 v-on:click="collapsed = !collapsed" class="expandable queue" v-bind:class="{'expanded' : !collapsed}">Queued vouchers</h1>
 
-        <message :text="queueStatus"></message>
+        <message :text="queueStatus || message.text" :state="message.state"></message>
 
         <button id="submitQueuedVouchers"
             class="cta queuedVouchers"
@@ -44,17 +44,21 @@
 
 <script>
 import Store from '../store.js';
+import mixin from '../mixins/mixins';
 import Message from './Message.vue';
 
 import constants from '../constants';
 
-const RESULT_TIMER = 5000;
+const RESULT_TIMER = 2000;
 
 export default {
     name: 'queue',
     components: {
         Message
     },
+    mixins: [
+        mixin.messages,
+    ],
     data: function() {
         return {
             queue: Store.queue,
@@ -65,7 +69,6 @@ export default {
             spinner: false,
             validate: false,
             fail: false,
-            message: '',
         }
     },
     watch: {
@@ -80,6 +83,9 @@ export default {
             },
             deep: true
         },
+        vouchers: function() {
+            this.setMessage(this.queueStatus, constants.MESSAGE_STATUS);
+        }
     },
     mounted: function() {
         if(Store.queue.sendingStatus) {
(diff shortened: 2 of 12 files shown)


src/components/Message.vue
```diff
@@ -0,0 +1,35 @@
+<template id="message">
+  <transition name="fade" v-if="visible">
+    <div class="message" v-bind:class="{'error': error, 'good': success}" v-html="text"></div>
+  </transition>
```
gingerbreadpup: maybe we could just call good success too? I think we need to get rid of the goods!

src/components/Queue.vue
```diff
@@ -70,23 +66,20 @@ export default {
...
-                  this.spinner = false;
-              }
-
-          },
-          deep: true
-      },
-
+        queue: {
+            handler: function(val) {
+                let queueState = val.sendingStatus || false;
```
gingerbreadpup: I have no idea but is let an ES5 thing?

gingerbreadpup (commented):
Looks great so far. :+1:

src/components/Queue.vue
ob6160: This was added in another PR. But seeing as we're making most things ES5 I guess we should change it to a var. Babel will convert it to a var either way.

src/components/Message.vue
ob6160: I agree, prefer success over good :grin:

gingerbreadpup (approved):
Looks good to me - Katja should probably approve too

src/components/Queue.vue
```diff
@@ -4,17 +4,7 @@
...
-                {{ message }}
-            </div>
-            <div v-if="message" class="message">
-                {{ message }}
-            </div>
-            <div v-else class="message">
-                [xXx]You have <strong>{{ vouchers.length }}</strong> voucher<span v-if="vouchers.length > 1">s</span> in your queue. Queued vouchers will be checked when you submit your queue.
-            </div>
-        </transition>
+        <message :text="message.text" :state="message.state"></message>
```
katjam: We have been explicitly using v-bind in other places.

src/components/Queue.vue
```diff
@@ -113,68 +128,77 @@ export default {
...
+            }
+
+            if (data.invalid.length === 1) {
+                invalid = "and 1 was invalid.";
+            } else {
+                invalid = "and " + data.invalid.length + " were invalid.";
+            }
+
+            var message
+                = "Thanks! Your queue has been successfully submitted. "
```
katjam, line 172: There was a ticket about not displaying for 0 values - but that might be another card?

src/constants.js
```diff
@@ -0,0 +1,5 @@
+export default {
```
katjam, line 1: Not sure what we gain from this abstraction.

src/pages/Scan.vue
```diff
@@ -7,9 +7,7 @@
                 <h1>Scan a voucher code</h1>
 
                 <form id="textVoucher" v-on:submit.prevent>
-                    <transition name="fade">
-                        <div v-if="errorMessage && (!showQueueMsg)" class="message error">{{ errorMessage }}</div>
-                    </transition>
+                    <message :text="message.text" :state="message.state"></message>
```
katjam: Explicit v-bind for consistency?

src/pages/Scan.vue
```diff
@@ -51,7 +49,7 @@
             </div>
 
             <div>
-                <queue @update="queueMessage"></queue>
+                <queue @message-update="setMessage"></queue>
```
katjam: Again - sometimes explicit v-on sometimes @

src/pages/Scan.vue
```diff
@@ -62,13 +60,20 @@
 <script>
 
 import Store from '../store.js';
+import mixin from '../mixins/mixins';
 import Profile from '../components/Profile.vue';
 import Queue from '../components/Queue.vue';
+import Message from '../components/Message.vue';
+
```
katjam: I bet there is a good reason for this blank line

src/components/Queue.vue
```diff
@@ -69,24 +69,29 @@ export default {
...
-                  this.spinner = false;
-              }
-
-          },
-          deep: true
-      },
-
+        queue: {
+            handler: function(val) {
+                var queueState = val.sendingStatus || false;
```
katjam: Will it be undefined if not exist?

src/components/Queue.vue
```diff
@@ -69,24 +69,29 @@ export default {
...
+                if(!queueState && val.sentData) {
+                    var message = this.genQueueSuccessMessage(val.sentData);
+                    this.emitMessage(message, constants.MESSAGE_SUCCESS);
+
```
katjam: I bet you have a good reason for this blank line.

src/components/Queue.vue
```diff
@@ -69,24 +69,29 @@ export default {
...
-                  this.showValidate();
-              } else if(!queueState) {
-                  this.spinner = false;
-              }
-
-          },
-          deep: true
-      },
-
+        queue: {
```
katjam, line 80: Documentation might be helpful so we know what is intended at a glance.

katjam (commented):
Just little niggles - probably ok to go in as is.

src/components/Queue.vue
ob6160: It'll still be falsey you're right.. I'll remove the false

src/components/Queue.vue
ob6160: I just took the message as I found it. It's not related directly to the message component so it should be resolved in another card most likely.

src/constants.js
ob6160: It makes it easier to remember what 0,1,2 mean. We could alternatively pass in "error", "success", "info" when we set the message components' type - but this felt cleaner to me. Plus as I mentioned earlier we could possibly place other things such as copy into this new constants store.

src/components/Queue.vue
ob6160: Added on line 80

katjam (approved):
Once Travis finishes. :hibiscus:

PR #170, opened 2017-07-13
1.1/441 account page improvements

gingerbreadpup:
A lot of css here to rework the accounts page. Also added a different icon, radio buttons, a new button at the bottom which is only enabled if a radio button is selected, informative copy at the top of the page. Hopefully these will make the page more usable - let me know if you have any other suggestions though.

https://trello.com/c/O7gGmrJ8/441-as-a-trader-i-need-to-understand-my-action-options-on-the-my-account-page

### DIFF
--- a/src/pages/Account.vue
+++ b/src/pages/Account.vue
@@ -13,10 +13,10 @@
                     <div class="tab thead">
                         <div class="row">
                             <div></div>
-                            <div>Request date</div>
+                            <div class="date"></div>
                             <div class="total"></div>
                             <div>Amount</div>
-                            <div>Email Record</div>
+                            <div class="select-record"></div>
                         </div>
                     </div>
 
@@ -27,7 +27,7 @@
                             <div> {{ payment.pended_on }}</div>
                             <div> {{ payment.vouchers.length }}</div>
                             <div class="amount">&pound;{{ payment.vouchers.length }}</div>
-                            <div class="email"><input type="radio" name="radio-group" value="Email this payment history record" required></div>
+                            <div class="email"><input type="radio" name="radio-group" value="Email this payment history record" required></label></div>
                         </div>
                         <div class="tab-content">
                             <div class="tab inner-thead">
--- a/src/pages/Account.vue
+++ b/src/pages/Account.vue
@@ -7,9 +7,13 @@
 
                 <transition name="fade"><div v-if="errorMessage" v-bind:class="[goodFeedback ? 'good message' : 'message' ]">{{ errorMessage }}</div></transition>
 
-                <p>Click the <span class="list-icon"><i class="fa fa-list" aria-hidden="true"></i></span> icon below to view a payment history record in more detail.</p>
-                <p>To email yourself a specific payment history record from the table below, select it and click 'Email selected payment history'.</p>
-                <p>Or, to email yourself all of your payment history, just click 'Email all payment history'.</p>
+                <div v-if="voucherPayments.length > 0">
+                    <p>Click the <span class="list-icon"><i class="fa fa-list" aria-hidden="true"></i></span> icon below to view a payment history record in more detail.</p>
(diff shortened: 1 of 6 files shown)


src/pages/Account.vue
```diff
@@ -7,38 +7,43 @@
 
                 <transition name="fade"><div v-if="errorMessage" v-bind:class="[goodFeedback ? 'good message' : 'message' ]">{{ errorMessage }}</div></transition>
 
+                <div v-if="voucherPayments.length > 0">
```
ob6160: This check will have to be changed a little because voucherPayments.length always returns 1. It's because Vue is attaching an Observer I believe.

ob6160: Probably something like this but in a computed value or even just inline:
`(voucherPayments[0]) ? voucherPayments[0].length > 0 : false`

ob6160 (dismissed):

src/pages/Account.vue
katjam: @gingerbreadpup I thought we already had the function defined - but ah ha to that's for vouchers length

src/pages/Account.vue
```diff
@@ -53,7 +58,10 @@
 
                 </div>
 
-                <button id="requestVoucherHistoryEmail" v-on:click="onRequestVoucherHistoryEmail">Email payment history</button>
+                <div class="cta-buttons">
+                    <button id="requestVoucherHistoryEmail" v-on:click="onRequestVoucherHistoryEmail">Email all payment history</button>
+                    <button v-on:click="onRequestSubmissionEmail" :disabled="selected">Email selected payment history</button>
```
katjam: Was it agreed with Kat to have the 2 buttons instead of just giving them one or other option? I think it makes the interface more confusing.

src/pages/Account.vue
gingerbreadpup: Are you sure? I've tested it and it seemed to function correctly.

tests/pages/account.test.js
```diff
@@ -75,12 +75,13 @@ test('Voucher total is equal to amount of vouchers in accordion', async t => {
...
         .click('button#continue')
+        .debug()
     ;
-    const voucherTotal = await el('label[for=tab-0] div').child(1).innerText;
+    const voucherTotal = await el('.count').innerText;
     expect(voucherTotal).to.equal('2');
 
     await t
-        .click('input[type=checkbox')
+        .click('input[type=checkbox]');
```
katjam: There's an extra ; here

katjam (commented):
Couple of bugfixes to go

src/pages/Account.vue
gingerbreadpup: Kat was on board with the two buttons. The button on the right for selected history is greyed out and disabled until a radio button is selected though, which I think helps.

src/pages/Account.vue
katjam: @ob6160 Maybe vue attaches the observer if you observe it but not in normal circumstances?

src/pages/Account.vue
katjam: Yeah - that would help.

src/pages/Account.vue
ob6160: Just tested with Steph and Alex - It's definitely returning an array with the array we want to check the length of inside of it.

ob6160: image

gingerbreadpup:
Ok. I think this is ready to go. I think @ob6160 is going to do the fix for the voucherPayments.length > 0 bit - not sure how you were planning to do it (in another PR or in here), but I'll leave it up to you.

gingerbreadpup:
Splicing fix works - will merge in once someone approves. Thanks!

katjam (approved):
Let's get this in!

PR #171, opened 2017-07-13
1.1/445 expandability improvements

gingerbreadpup:
Queue section: 'Queued vouchers' is now just a heading. Below it is a list icon and the text 'view your queued vouchers', which shows and hides the queued vouchers.

Payment page: Same as above, the heading is just a heading and below it is a show/hide list icon and text which says 'View your voucher list'.

These changes, along with the PR for the account page, should make the expandability functionality consistent and more obvious across the app. Card [here.](https://trello.com/c/KvcFPqKa/445-make-expandability-function-more-obvious-and-consistent)

### DIFF
--- a/src/components/Queue.vue
+++ b/src/components/Queue.vue
@@ -2,7 +2,7 @@
     <transition name="fade"  v-if="currentlyShown">
     <div class="content narrow queuedVouchers">
 
-        <h1 v-on:click="collapsed = !collapsed" class="expandable queue" v-bind:class="{'expanded' : !collapsed}">Queued vouchers</h1>
+        <h1>Queued vouchers</h1>
 
         <transition name="fade" v-if="show">
             <div v-if="fail && message" class="message error">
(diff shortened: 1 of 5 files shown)


src/components/Queue.vue
```diff
@@ -16,6 +16,8 @@
             </div>
         </transition>
 
+        <div v-on:click="collapsed = !collapsed" class="expandable queue" v-bind:class="{'expanded' : !collapsed}"><i class="fa fa-list" aria-hidden="true"></i></div>
```
katjam, line 19: As per my comment to @ob6160 we seem to sometimes be using v-on and v-bind and sometimes using @ and :

Either fine - but should be consistent. I think we are mostly being explicit.

src/sass/app.scss
```diff
@@ -370,21 +370,30 @@ header {
...
   }
-  &.expanded:before {
-    font-weight: bold;
-    font-family: FontAwesome;
-    content: "\f0d7";
+  .fa {
+    font-size: 1rem;
     padding-right: 10px;
   }
+  &:after {
```
katjam, line 406: Maybe also add a test for this new feature?

katjam (approved):
Looks good. could use a bit more test coverage but...

PR #172, opened 2017-07-14
Send selected date to the voucher-history-email endpoint

ob6160:


src/pages/Account.vue
```diff
@@ -83,18 +83,24 @@
...
         created: function () {
 
         },
         methods: {
+            recordSelect: function(event) {
+                this.selected = false;
+                // Default to requesting all.
+                this.selectedDate = event.target.id || null;
+            },
             onRequestSubmissionEmail : function(event) {
```
katjam: Do we still need the event here?

ob6160: nope!

katjam (commented):
I think this is good to go now.. but maybe just check that the API expects null and not ''

src/pages/Account.vue
```diff
@@ -83,18 +83,24 @@
                 voucherPayments: Store.trader.pendedVouchers,
                 errorMessage : Store.error,
                 goodFeedback : false,
-                selected : true
+                selected : true,
+                selectedDate: null
```
katjam, line 87: I think it's fine - but can't remember what the API expects.

ob6160: I just checked - it looks like null

katjam: :+1:

katjam (approved):
Do this thing! :boat:

PR #173, opened 2017-07-14
1.1/441 account ui

gingerbreadpup:
Tweaks. Hopefully done with the account page for now (except for adding the confirmation message which is part of another card).


katjam (approved):
:koala:

PR #174, opened 2017-07-14
1.1/185 request payment feedback

gingerbreadpup:
I branched off @ob6160's messaging component branch, which he has since merged in - hence all the commit gubbins below. I could just pull from 1.1 and make a new branch and redo this, since there's not much code for this ticket.

But anyway this works now, error message on the payment page if offline, and redirect to the accounts page with success message if successful.

### DIFF
--- a/src/pages/Account.vue
+++ b/src/pages/Account.vue
@@ -81,9 +81,6 @@
                 goodFeedback : false
             }
         },
-        created: function () {
-
-        },
         methods: {
             onRequestSubmissionEmail : function(event) {
                 var url = '/traders/' + Store.trader.id + '/voucher-history-email';
--- a/src/pages/Account.vue
+++ b/src/pages/Account.vue
@@ -130,6 +130,11 @@
         },
         mounted: function () {
             Store.getVoucherPaymentState();
+            // TODO: Have a standard way of having global router messages.
+            var message = this.$router.message;
+            if(message) {
+                this.setMessage(message.text, message.state);
+            }
         }
     }
 </script>
--- a/src/pages/Account.vue
+++ b/src/pages/Account.vue
@@ -5,7 +5,9 @@
             <div class="content fullwidth">
                 <h1>Requested Payments</h1>
 
-                <transition name="fade"><div v-if="errorMessage" v-bind:class="[goodFeedback ? 'good message' : 'message' ]">{{ errorMessage }}</div></transition>
+                <transition name="fade">
+                    <message :text="message.text" :state="message.state"></message>
+                </transition>
 
                 <div class="accordion">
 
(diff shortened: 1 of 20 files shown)


src/pages/Account.vue
```diff
@@ -146,6 +146,11 @@
         },
         mounted: function () {
             Store.getVoucherPaymentState();
+            // TODO: Have a standard way of having global router messages.
```
ob6160, line 149: We should ticket this but put it low down in the list or maybe just tech debt. I would imagine that we would expand the message mixin to include some functionality which lets you set a message on the router. It only really becomes something we'll need to do if we want to send messages over the router in other places apart from here.

src/pages/Payment.vue
```diff
@@ -39,6 +39,10 @@
                     </div>
 
                     <instructions></instructions>
+                    
+                    <transition name="fade">
+                        <message :text="message.text" :state="message.state"></message>
```
ob6160: I think we're using the v-bind syntax for those two props from now on.

src/pages/Payment.vue
```diff
@@ -77,15 +89,23 @@ export default {
...
             Store.pendRecVouchers(
                 // on Success, route to /account
                 function(response) {
-                    this.$router.push('/account');
+                    this.showConfirmation();
                 }.bind(this),
                 // on Failure... hook for an alert?
                 function(error) {
-                }
+                    if (!Store.netMgr.online) {
```
ob6160: How come we have to be offline for the error message to show here. Do we want a message to show if it errors whilst we're online?

src/pages/Scan.vue
```diff
@@ -49,7 +51,7 @@
             </div>
 
             <div>
-                <queue v-on:message-update="setMessage"></queue>
+                <queue v-on:message-update="setMessage"></queue>
```
ob6160: Spaces ✂️ 🙈

ob6160 (changes requested):
Looks good, just a few questions and one or two changes.

src/pages/Payment.vue
gingerbreadpup: At the moment the only reason it would error is if it's offline - right? Also it wasn't working for me without this code, so I figured we needed it.

ob6160: If the error callback only fires when we're offline why do we need to check again if we're offline before we show the message? I'd also imagine that the error callback would fire if we're online and the server suffers an error for some reason.

gingerbreadpup: Ah, didn't realise it was only firing when offline - was just going by how it's done on other pages as far as I could tell. Will get rid of it and check I can get the error message to display.

ob6160: That sounds like a good plan 😄 I think we do it on Tap and Scan because we need to make sure that we're offline before we show queue specific messages.

gingerbreadpup: Hmm - just pushed up a commit reverting the error bit to how it was - but now I can't get an error message to show if I'm offline. Anyway... time for bed...

ob6160: You just missed the .bind(this), I've pushed the fix and a small wording change.

ob6160 (approved):

PR #175, opened 2017-07-17
1.1/441 tweaks

gingerbreadpup:
Removed the instructions component, and have put the static payment instruction text on the end of the success message on the account page after payment has been requested.


alexalala (approved):
:shipit: Looks good!

src/pages/Account.vue
```diff
@@ -12,9 +12,8 @@
                 <div v-if="voucherPayments.length > 0">
 
                     <div>
-                        <p>[xXx]Click the <span class="list-icon"><i class="fa fa-list" aria-hidden="true"></i></span> icon below to view a payment record in more detail.</p>
-                        <p>[xXx]To email yourself a specific payment record from the table below, select it and click 'Email selected payment record'.</p>
-                        <p>[xXx]Or, to email yourself all of your payment records, just click 'Email all payment records'.</p>
+                        <p>Click the <span class="list-icon"><i class="fa fa-list" aria-hidden="true"></i></span> icon below to view a payment record in more detail.</p>
+                        <p>To email yourself a specific payment record from the table below, select it and click 'Email selected payment record'.</p>
```
ob6160, line 16: Was "Or, to email yourself all of your payment records, just click 'Email all payment records'." meant to make it too?

src/pages/Payment.vue
```diff
@@ -90,7 +86,7 @@ export default {
     },
     methods: {
         showConfirmation: function() {
-            this.setMessage("Thanks, your payment request has been sent.", constants.MESSAGE_SUCCESS);
+            this.setMessage("Thanks, your payment request has been sent. Please take your vouchers to your market representative for them to be sent off.", constants.MESSAGE_SUCCESS);
```
katjam, line 89: I think we still need it to be a variable/ component though - because different markets have different instructions. See: https://trello.com/c/VZF4RYMN/185-add-market-specific-variations-for-pay-claim-voucher-return-instructions

katjam (changes requested):
I guess we can put back in with https://trello.com/c/VZF4RYMN/185-add-market-specific-variations-for-pay-claim-voucher-return-instructions or do it differently - but the plan will be to get the instructions from the market object through the trader.

alexalala (changes requested):
Amended approval due to everyones comments

src/pages/Payment.vue
gingerbreadpup: Yes, I think it will need to be a variable - but I believe that's part of the card @ob6160 is working on, so I thought we could merge this in and then he could replace the static text with a variable.

src/pages/Account.vue
katjam: What he said.

src/pages/Payment.vue
katjam: I guess depends on if we will need to show instructions in the message - and by request - (help or payment instructions) or if it is only in this place.

ob6160: I agree.

src/pages/Account.vue
gingerbreadpup: Yep, I left this bit out intentionally because I think there's a lot of copy above the table and I think that (hopefully) the email all button doesn't actually need explanation.

src/pages/Payment.vue
katjam: Okeedokee - I agree with @alexalala then. :+1: :zap:

katjam (approved):
Yahoo! :cow:

alexalala (approved):
Re-approved!!!!!!

src/pages/Payment.vue
gingerbreadpup: I believe that this is the only place we'll be showing instructions, we hadn't planned to put them anywhere else as far as I know.

gingerbreadpup:
Thank you all so much :stuck_out_tongue:

PR #179, opened 2017-07-18
Set deploy branch to 1.1/release and add file to gitignore

ob6160:


.gitignore
```diff
@@ -5,3 +5,4 @@ npm-debug.log
 yarn-error.log
 screenshots
 .idea
+package-lock.json
```
katjam: Why have you added and ignored package-lock? If there is one, we'd want to include it.

katjam (changes requested):

.gitignore
ob6160: I didn't add it - npm added it itself at some point. I'll remove the ignore and we can make a ticket about including it.

katjam: Did you run an update?

ob6160: Possibly. I can't remember though.

katjam (approved):
:goat: But DON'T merge until after filming.

PR #180, opened 2017-07-18
Add new /market endpoint and get payment success message from new mar…

ob6160:
…ket object.

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -10,7 +10,8 @@ var store = {
         id: null,
         pendedVouchers: [],
         vouchers: [],
-        recVouchers: []
+        recVouchers: [],
+        market: {},
     },
     netMgr: NetMgr,
     auth: false,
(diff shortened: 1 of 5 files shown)


ob6160:
This also adds a fix for the router message. Previously it wasn't being cleared after the redirect meaning that it persisted even when the user switched pages / traders.

src/store.js
```diff
@@ -206,6 +209,18 @@ store.getVoucherPaymentState = function () {
     return true;
 };
 
+store.getMarketInfo = function() {
+    this.netMgr.apiGet('/traders/' + this.trader.id + '/market', function(response) {
+        if(response.data) {
+            for(var i in response.data) {
+                this.trader.market[i] = response.data[i];
+            }
+
```
gingerbreadpup: Could lose this empty line?

gingerbreadpup (approved):
This all looks good to me :+1:

PR #181, opened 2017-07-19
1.1/185 add market specific variations for pay claim voucher return instructions

ob6160:


gingerbreadpup (approved):
:shipit:

ob6160:
I'll hold off merging until the service change is in.

src/constants.js
```diff
@@ -2,5 +2,9 @@ export default {
     MESSAGE_ERROR: 0,
     MESSAGE_STATUS: 1,
     MESSAGE_SUCCESS: 2,
-    MESSAGE_WARNING: 3
+    MESSAGE_WARNING: 3,
+    copy: {
+        PAYMENT_REQUEST_DEFAULT: "Thanks, your payment request has been sent. Please take your vouchers to your market representative for them to be sent off.",
```
charlesstrange2, line 7: Has this been approved by Steph - i presume "yes" becuase it's copied from elsewhere but check just in case? otherwise needs [xXx]

src/constants.js
```diff
@@ -2,5 +2,9 @@ export default {
...
+        PAYMENT_REQUEST_ERROR: "There was a problem with your payment request, please try again later.",
```
charlesstrange2, line 8: See above. Also, i remember Katja saying in scoping that she'd prefer if we used the server-passed error for error messages rather than  having the UI layer reinterpret them. There may be another card for that.

charlesstrange2 (commented):
code cool, notes about errors.

src/constants.js
ob6160: Yep it didn't have an [xXx] when I copied it.

charlesstrange2 (approved):
:shipit:

src/constants.js
ob6160: I agree - the server should specify the error message, as it was like this before (I just moved into constants) we should definitely fix this in another card.

PR #182, opened 2017-07-19
1.1/416 as trader i would like to see a default prefix relevant to my area so that my voucher code entry is faster

ob6160:
I believe that it really is this simple. It will default to an empty string if for some reason the market object isn't included in the trader request (which shouldn't happen anyway).


gingerbreadpup (approved):
:beach_umbrella:

charlesstrange2 (approved):
Cool :cool:
:shipit:

PR #183, opened 2017-07-21
1.1/271 dupe message

katjam:


src/components/Queue.vue
```diff
@@ -83,7 +83,7 @@ export default {
                 // so that we can reflect any changes in the Queue component.
                 var queueState = val.sendingStatus;
                 if(!queueState && val.sentData) {
-                    var message = this.genQueueSuccessMessage(val.sentData);
+                    var message = this.queueSuccessMessage(val.sentData);
```
ob6160, line 86: Do we really need this method anymore if it's just returning val.sentData.message?

ob6160 (changes requested):

src/components/Queue.vue
katjam: Nope but we are doing a refactor. So taking one step at a time. These are the final hours!

ob6160: ok! :)

src/pages/Scan.vue
```diff
@@ -128,17 +121,15 @@ export default {
...
                         if (!Store.netMgr.online) {
                             this.showQueued();
-                            this.setMessage("Not enough signal, voucher queued.", constants.MESSAGE_ERROR);
+                            this.setMessage("Not enough signal, voucher queued.", constants.MESSAGE_WARNING);
                         }
                     }.bind(this));
 
                 // Do anyway.
                 this.voucherCode = "";
-                this.sponsorCode = "";
```
ob6160, line 137: Why were these two lines removed? They're causing some of the tests to fail I think.

ob6160:
This doesn't seem to be fully working - closing - will fix.

src/pages/Scan.vue
katjam: Not sure. @alexalala Do you think that was a mistake?

src/pages/Scan.vue
```diff
@@ -99,20 +100,12 @@ export default {
...
-                                this.setMessage("Please enter a valid voucher code.", constants.MESSAGE_ERROR);
-
-                            } else if (data.fail.length > 0) {
-                                this.showFail();
-                                this.setMessage("It looks like this code has already been added, please double check and try again. If you are still unable to add the voucher code, don\'t worry - you will still receive payment if you send it in with your other vouchers.", constants.MESSAGE_WARNING);
-                            }
-
+                        if (response.error) {
+                            this.showFail();
+                            this.setMessage(response.error, constants.MESSAGE_ERROR);
```
ob6160, line 105: Also the messages are part of the "response.data" object not the response itself.

src/pages/Scan.vue
alexalala: @katjam @ob6160 quite possibly!!! Add them back in if they need to be :)

ob6160: They're pretty essential to the functioning of the scan page afaik.

src/pages/Scan.vue
```diff
@@ -128,17 +121,15 @@ export default {
...
 
                 // Do anyway.
                 this.voucherCode = "";
-                this.sponsorCode = "";
-                this.$refs.sponsorBox.focus();
             } else {
-              this.showFail();
-              this.setMessage("Please enter a valid voucher code.", constants.MESSAGE_ERROR);
+                this.showFail();
+                this.setMessage(response.error, constants.MESSAGE_ERROR);
```
ob6160, line 132: There is no response in scope here.

src/pages/Scan.vue
alexalala: @ob6160 yeah it was probably a mistake. As I said, add them back in if you need to

PR #184, opened 2017-07-21
Fix duplicate scan submission and allow all size voucher codes to be subm…

ob6160:
…itted when carriage return is entered.


charlesstrange2 (approved):
Looks ok.
:shipit:

PR #189, opened 2017-07-25
.htaccess solution

charlesstrange2:
We serve staging and production off Apache; We dev on Node. Thus i gad to log into staging, put the file in the root directory (emulating webpack copying it to /dist) and check it was being correctly applied.


ob6160 (approved):
Those rewrite conditions seem sensible :shipit:

katjam (approved):

PR #195, opened 2017-08-01
1.1/148 redo auth service to talk to backend

ob6160:
It might be a good idea to test this on your own machines as well as just reviewing the code as there might be unforeseen consequences that haven't appeared on my PC.


ob6160:
Needs a bit more work.

PR #196, opened 2017-08-02
Add http so link opens correctly

charlesstrange2:


ob6160 (approved):
Looks good. Do they not offer an https version?

PR #208, opened 2020-05-12
1.3/1533 upgrade 3rd party

katjam:
- update 3rd party deps
- fix tests to be compatible with latest testcafe


AlinChindea (approved):
👑 🐝

PR #209, opened 2020-05-14
1.3/1580 asset management bug manifest and logo

katjam:
https://trello.com/c/W9WZJp5M/1580-is-the-webpack-build-bundle-working-noticed-that-logo-not-rendering-in-dev

- fix file load image settings for logo
- generate manifest with webpack


katjam:
This seems to work. I basically copied the whole existing manifest, we might want to do something with the icons that is more dynamic - but not worth time until it is an issue.

webpack.config.js
```diff
@@ -45,7 +46,46 @@ module.exports = {
...
+                    src: path.resolve('src/assets/launcher-192x192.png'),
+                    size: '192x192',
+                },
+                {
+                    src: path.resolve('src/assets/icon.svg'),
+                    size: '193x193',
+                },
+                {
+                    src: path.resolve('src/assets/launcher-512x512.png'),
+                    size: '512x512',
```
charlesstrange2, line 84: I wonder if we need more/new/reformatted Icons for the PWA standard - both play and istore have changed their standards, has the PWA changed too?

charlesstrange2 (approved):
give it a spin! 🌪️

webpack.config.js
katjam: Yes - we probably do, but I don't think it's a priority. This plugin seems to be able to generate sizes from one icon and an array of sizes... but I thought I'd leave that for a day in the distant future.

I'll note an AC on the card for @jermyk to test the icon for all supported devices.

PR #211, opened 2020-05-28
1.3/1598 bug build version is old

charlesstrange2:
https://trello.com/c/oPeB11hu/1598-bug-build-version-is-old


AlinChindea (approved):
🦖 🎉

PR #216, opened 2020-12-07
1.4/1654 rollup market repo vvv e

charlesstrange2:
https://trello.com/c/Byyye9LH/1654-rollup-market-repo-vvv-e
Much fiddling to get the old packages up to date.


package.json
```diff
@@ -1,49 +1,49 @@
...
+    },
+    "licence": "UNLICENSED",
+    "name": "arcvmarket",
+    "private": true,
+    "scripts": {
+        "dev": "cross-env NODE_ENV=development webpack-dev-server --open --hot",
+        "build": "cross-env NODE_ENV=production webpack --mode production --progress --hide-modules --display-error-details",
+        "test": "testcafe chrome,firefox ./tests/*/*.test.js --app 'cross-env NODE_ENV=development webpack-dev-server' --app-init-delay 10000"
+    },
+    "version": "1.4.0-dev"
```
AlinChindea, line 48: not 1.4.0?

src/config.js
```diff
@@ -2,7 +2,7 @@
 // --- Defaults ---
 var env = 'production',
     apiBase = 'https://voucher-admin.alexandrarose.org.uk/api',
-    appVersion = 'v1.3.0';
+    appVersion = 'v1.4.0-dev';
```
AlinChindea, line 5: not v1.4.0?

src/main.js
```diff
@@ -60,34 +60,66 @@ const router = new VueRouter({
...
-router.beforeEach((to, from, next) => {
-    var auth = Store.netMgr.isAuth();
+// Nasty hack to get round the new NavigationErrors
+const originalPush = VueRouter.prototype.push
+VueRouter.prototype.push = function push(
+    location,
+    onResolve,
+    onReject
+) {
+    console.log("running error")
```
AlinChindea: don't forget this in

charlesstrange2: oh, pants! 👖

webpack.config.js
```diff
@@ -1,23 +1,20 @@
...
+const HtmlWebpackPlugin = require('html-webpack-plugin');
+const CopyWebpackPlugin = require('copy-webpack-plugin');
+const OfflinePlugin = require('offline-plugin');
+const TerserPlugin = require('terser-webpack-plugin');
+const { VueLoaderPlugin } = require('vue-loader');
+const WebpackPwaManifest = require('webpack-pwa-manifest');
+const gitRevisionPlugin = new GitRevisionPlugin();
 
 module.exports = {
+    mode: 'none',
```
AlinChindea, line 14: curious what mode: 'none' does here

charlesstrange2: Ok, so, "mode" needs to exist to to tell webpack if we're prod or dev. However, we set those by variables in the scripts section, so it's (i think) redundant here. hence "none"

It changes how webpack applies certain optimisations.

AlinChindea: I know about mode, just haven't seen the property 'none'. if you skip it, default is mode: 'production', so you don't want to do that I think

charlesstrange2:
Travis is slow

src/main.js
```diff
@@ -60,34 +60,66 @@ const router = new VueRouter({
         join('/')+'/'
 });
 
-// Route Guard rules for directing users
-router.beforeEach((to, from, next) => {
-    var auth = Store.netMgr.isAuth();
+// Nasty hack to get round the new NavigationErrors
+const originalPush = VueRouter.prototype.push
+VueRouter.prototype.push = function push(
```
AlinChindea, line 65: aren't we duplicating `VueRouter.prototype.push`?

charlesstrange2: 😬 
yes, so, Vue-Router 3.0 has decided to throw an exception if you
- have a page that programmatically redirects
- navigates back to the page we're on

There was a lot of whinging about this from the vue router users. It's do do with them quietly shifting to a promise based system rather than a callback based system. 

The official solution is to 
```
.catch(function() { ... deal with errors here ... })
``` 

after every `router.push()` ...
which this function effectively tries to do.

It's not a long term solution, but it does get us running

PR #218, opened 2020-12-09
1.4/1699 exp add linting

charlesstrange2:
OK, so this is the result of linting and prettying an entire project at once

https://trello.com/c/nSRDDFvA/1669-add-linting-prettier-on-commit-ee

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
         "eslint": "^7.15.0",
         "eslint-config-prettier": "^7.0.0",
         "eslint-plugin-prettier": "^3.2.0",
+        "eslint-plugin-standard": "^5.0.0",
         "eslint-plugin-vue": "^7.2.0",
         "file-loader": "^6.2.0",
         "font-awesome": "^4.7.0",
--- a/package.json
+++ b/package.json
@@ -30,6 +30,8 @@
         "font-awesome": "^4.7.0",
         "git-revision-webpack-plugin": "^3.0.6",
         "html-webpack-plugin": "^4.5.0",
+        "husky": "^4.3.5",
+        "lint-staged": "^10.5.3",
         "node-sass": "^4.14.1",
         "offline-plugin": "^5.0.7",
         "prettier": "^2.2.1",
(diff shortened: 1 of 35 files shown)


package.json
```diff
@@ -43,7 +52,21 @@
     "scripts": {
         "dev": "cross-env NODE_ENV=development webpack-dev-server --open --hot",
         "build": "cross-env NODE_ENV=production webpack --mode production --progress --hide-modules --display-error-details",
-        "test": "testcafe chrome,firefox ./tests/*/*.test.js --app 'cross-env NODE_ENV=development webpack-dev-server' --app-init-delay 10000"
+        "test": "testcafe chrome,firefox ./tests/*/*.test.js --app 'cross-env NODE_ENV=development webpack-dev-server' --app-init-delay 10000",
+        "lint": "eslint --ext .js,.vue src tests",
+        "lint-autofix": "eslint --ext .js,.vue src tests --fix",
```
katjam, line 57: Do you need the `.` to specify all recursively? https://eslint.org/docs/user-guide/command-line-interface#-ext

Also - have found that if you lint on commit, then people can manually lint with eslit directly and don't need to specify all these as scripts.

If you do keep them, they should be in sync with your staged command (see below)

package.json
```diff
@@ -43,7 +52,21 @@
...
-    "version": "1.4.0-dev"
+    "version": "1.4.0-dev",
+    "husky": {
+        "hooks": {
+            "pre-commit": "lint-staged"
+        }
+    },
+    "lint-staged": {
+        "*.{js,json,yaml,json,vue}": [
+            "eslint --ext .js,.vue src tests --fix --quiet --cache"
```
katjam, line 69: Should this be `npm run lint-autofix` ?

katjam (commented):
Much improved. A couple comments.

katjam (commented):
I'm sad about the so many unused imports, but good they can be tidied now.

package.json
charlesstrange2: these lint commands where there to lint and fix manuallywhile i was kicking the tires. remove?
the staged command has quiet and cache only flags to only lint the current changed files.

package.json
charlesstrange2: needs the quiet and cache flags to only check then changed files and not make a verbose fuss.

charlesstrange2:
I've resolved the outstanding complaints that the vue/essential rules had
- fixed the missing keys
- unwound the impure functions.

charlesstrange2:
closing to recreate new branch

PR #221, opened 2020-12-11
1.4/1653 update links for policies in the app

AlinChindea:
Trello 🎫 [https://trello.com/c/GdjPZpbX/1653-update-links-for-policies-in-the-app-vvv-ee](https://trello.com/c/GdjPZpbX/1653-update-links-for-policies-in-the-app-vvv-ee)


tests/pages/login.test.js
```diff
@@ -34,7 +49,7 @@ test("Can log in", async (t) => {
     expect(pagePath).to.equal(url + "/");
 });
 
-test("Can log out", async (t) => {
+test.only("Can log out", async (t) => {
```
AlinChindea:
```suggestion
test("Can log out", async (t) => {
```

src/App.vue
```diff
@@ -2,16 +2,25 @@
...
 export default {
     name: "app",
     components: {
         Masthead,
+        PrivacyLink
     },
+    computed: {
+        footerShow: function () {
+            const paths = ["/", "/scan"];
+            return !(paths.indexOf(this.$route.path) === -1);
```
charlesstrange2, line 21: nb.
if we ever to es2016, it has `[].inlcudes()`

tests/pages/login.test.js
```diff
@@ -46,6 +60,8 @@ test("Can log out", async (t) => {
     await t.click(logout);
     const pagePath = await t.eval(() => document.documentURI);
     expect(pagePath).to.equal(url + "/login");
+    const privacyLink = await el("#privacy").exists;
```
charlesstrange2, line 63: a little cheeky, hijacking the logout page to see the privacy policy; 
might have been better to have an "i can see the privacy policy link on the logout page" test

charlesstrange2 (approved):
I think I'd have preferred the logout page privacy link to be a separate test but it works

src/App.vue
AlinChindea: not supported by IE - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes#Browser_compatibility

tests/pages/login.test.js
AlinChindea: wouldn't that just duplicate teh whole `Can log out`test and just append two more lines of code? the ones I've added here. Also, when you log out you're redirected to the login page which already contains separate tests about the privacy link. come to think about it, it's redundant to add it even in the "Can log out" test as long as we redirect to login on logout.

PR #223, opened 2020-12-16
1.4/1677 change what icons are shown when I am forbidden to tap

AlinChindea:
### Trello 🎫 [https://trello.com/c/vTZpX0zv/1677-change-what-icons-are-shown-when-i-am-forbidden-to-tap-vv-e](https://trello.com/c/vTZpX0zv/1677-change-what-icons-are-shown-when-i-am-forbidden-to-tap-vv-e)

### DIFF
--- a/src/components/Toolbar.vue
+++ b/src/components/Toolbar.vue
@@ -11,7 +11,7 @@
                 >
             </div>
         </transition>
-        <div class="input-icons">
+        <div class="input-icons" v-if="showIcon">
             <router-link v-bind:to="'/'"
                 ><i class="fa fa-keyboard-o" title="Type a voucher code"
                     ><span class="hidden offscreen"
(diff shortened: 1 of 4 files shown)


src/components/Toolbar.vue
```diff
@@ -11,7 +11,7 @@
                 >
             </div>
         </transition>
-        <div class="input-icons">
+        <div class="input-icons" v-if="showIcon">
```
AlinChindea, line 14: @charlesstrange2 Should we use `v-show` instead? From the docs: https://vuejs.org/v2/guide/conditional.html#v-if-vs-v-show
```
Generally speaking, v-if has higher toggle costs while v-show has higher initial render costs. So prefer v-show if you need to toggle something very often, and prefer v-if if the condition is unlikely to change at runtime.
```

charlesstrange2: I'd keep v-if.
v-show hides with css (iirc), and i think it's better if this never gets rendered to dom

charlesstrange2 (approved):
🥇

PR #226, opened 2021-01-05
created a user booting function; added message passing to the login

charlesstrange2:
https://trello.com/c/SBHNJW0M/1690-market-is-not-logging-people-out-when-logged-in-at-time-in-trader-disabling

NetMgr creates an event when it gets a 403, from someone misbehaving when connecting to an API endpoint they shouldn't
Main picks that up and logs the user out before sending them on to the Login page, with a message code that renders an appropriate  message.

### DIFF
--- a/src/components/Profile.vue
+++ b/src/components/Profile.vue
@@ -28,6 +28,7 @@ export default {
     methods: {
         onLogout: function () {
             Store.unAuthenticate();
+            // Artificially using a HTTP code to set to 200 for "Ok, I did that"
             const routeObj = { name: "login", params: { logoutReason: 200 } };
             this.$router.push(routeObj);
         },
--- a/src/components/Profile.vue
+++ b/src/components/Profile.vue
@@ -28,7 +28,8 @@ export default {
     methods: {
         onLogout: function () {
             Store.unAuthenticate();
-            this.$router.push("/login");
+            const routeObj = { name: "login", params: { logoutReason: 200 } };
+            this.$router.push(routeObj);
         },
     },
     mounted: function () {
--- a/src/constants.js
+++ b/src/constants.js
@@ -17,5 +17,12 @@ export default {
             " has been successfully removed from your list.",
         DELETE_VOUCHER_FAIL:
             " cannot be removed at this time, please try later.",
+        FORCED_LOGOUT:
+            "The system has logged you out, please login again in a little while.",
+        USER_LOGOUT: "You have logged out.",
+        TIMEOUT_LOGOUT: "Your session has timed out, please login again.",
+        UNKNOWN_EVENT: "Something unusual has happened.",
+        INVALID_CREDENTIALS:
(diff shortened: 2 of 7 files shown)


src/components/Profile.vue
```diff
@@ -28,7 +28,8 @@ export default {
     methods: {
         onLogout: function () {
             Store.unAuthenticate();
-            this.$router.push("/login");
+            const routeObj = { name: "login", params: { logoutReason: 200 } };
```
katjam, line 32: what does logoutReason 200 mean? Could be made more clear here that it means "normal logout"

src/constants.js
```diff
@@ -17,5 +17,12 @@ export default {
             " has been successfully removed from your list.",
         DELETE_VOUCHER_FAIL:
             " cannot be removed at this time, please try later.",
+        FORCED_LOGOUT:
+            "The system has logged you out, please login again in a little while.",
+        USER_LOGOUT: "You have logged out.",
+        TIMEOUT_LOGOUT: "Your session has timed out, please login again.",
+        UNKNOWN_EVENT: "Something unusual has happened.",
```
katjam, line 24: :)

src/main.js
```diff
@@ -37,23 +35,29 @@ const routes = [
...
-        }
+        },
     },
     { path: "/account", component: Account, meta: { auth: true } },
     { path: "/scan", component: Scan, meta: { auth: true } },
     { path: "/payment", component: Payment, meta: { auth: true } },
-    { path: "/login", component: Login, meta: { auth: false } },
+    {
+        path: "/login",
+        name: "login",
```
katjam, line 56: The others don't have names. Does it matter?

src/pages/Login.vue
```diff
@@ -80,6 +80,32 @@ export default {
             env: Config.env,
         };
     },
+    mounted: function () {
```
katjam, line 83: Nice!

src/services/netMgr.js
```diff
@@ -252,6 +252,12 @@ NetMgr.axiosInstance.interceptors.response.use(
             NetMgr.setOnlineStatus(false);
         }
 
+        // is it a 403? USer tried something bad, broadcast a Logout event, someone will deal.
```
katjam: Typo 
```suggestion
        // is it a 403? User tried something bad, broadcast a Logout event, someone will deal.
```

katjam (approved):
A tiny typo and maybe some extra doc needed on standard logout to ref the logout reason Consts but otherwise ace!

src/components/Profile.vue
charlesstrange2: I'll add a comment

src/main.js
charlesstrange2: We call them by path elsewhere; oddly we need a name to do this, apparently 🤷

PR #228, opened 2021-01-06
1.4/1695 bug lint staged seems to check everything

charlesstrange2:
https://trello.com/c/tuist2Pf/1695-bug-lint-staged-seems-to-check-everything-rather-than-staged

OK, so I was forcing prettier to check all things in the lint-staged section of `package.json`.
This PR should also push Axios up a point release to deal with a fresh vulnerability.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -10,7 +10,7 @@ module.exports = {
     env: {
         browser: true,
     },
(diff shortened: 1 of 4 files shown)


AlinChindea (approved):
🏇 take it for a ride!

PR #229, opened 2021-01-07
refactor the login page message system ...

charlesstrange2:
... to be a more generic than just reacting to the logout event.

OK, so to show a nice message on the login page when we route back to reset a password, i had  to move some of the logic about.


AlinChindea (approved):
🐸 peachy!

PR #255, opened 2022-02-22
Disable requestPayment button on press.

neonleann:
https://trello.com/c/vQSbo7hV/1752-trader-scanning-multiple-submits-vv-ee

### DIFF
--- a/src/pages/Payment.vue
+++ b/src/pages/Payment.vue
@@ -132,6 +132,9 @@ export default {
             this.$router.push("/account");
         },
         onRequestPayment() {
+            document
+                .getElementById("requestPayment")
+                .setAttribute("disabled", "disabled");
             Store.pendRecVouchers(
                 // on Success, route to /account
                 function () {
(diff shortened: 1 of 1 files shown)


src/pages/Payment.vue
```diff
@@ -132,6 +132,9 @@ export default {
             this.$router.push("/account");
         },
         onRequestPayment() {
+            document
+                .getElementById("requestPayment")
+                .setAttribute("disabled", "disabled");
```
charlesstrange2, line 137: check this doesn't remain disabled. after an error [assumption, they might want to try again]; we might do this by sticking a line in the second function that pendRecVouchers takes, which is it's "failure" function.

charlesstrange2 (approved):

src/pages/Payment.vue
neonleann: Done.

charlesstrange2 (approved):
hotfixed! 🤞

PR #257, opened 2022-03-07
1.5/1762 in market we only show the most recent payment request for outstanding vouchers

charlesstrange2:
When the accounts page is mounted it fetches the voucher payement state
That now should have link headers in.

So we add the parseing of link headers inside store::getVoucherPaymentState
- here we update the central store,
- which is watched by the account page template.

if there are more than one page of results, then we draw the pagination controls
In the pagination controls we detect if we need to disable the buttons in a first, prev, next, last pattern
and also draw the "x / y"  page section

https://trello.com/c/UqyVMi3p/1651-trader-has-6mb-payment-history-which-refuses-to-load-in-app-vvv-eee

### DIFF
--- a/src/pages/Account.vue
+++ b/src/pages/Account.vue
@@ -104,10 +104,13 @@
                         <div
                             v-if="!pgBtnIsDisabled('current')"
                             class="tab row-pagination"
+                            aria-label="Pagination"
+                            aria-describedby="pagination-label"
                         >
                             <div>
                                 <button
                                     id="first"
+                                    aria-label="First Page"
                                     class="small-button"
                                     @click="pgChangePage"
                                     :disabled="pgBtnIsDisabled('first')"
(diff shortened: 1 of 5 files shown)


neonleann (approved):
I gave it a try and it works a treat. Lovely pagination buttons🥇!

src/pages/Account.vue
```diff
@@ -100,6 +100,55 @@
...
+                            class="tab row-pagination"
+                        >
+                            <div>
+                                <button
+                                    id="first"
+                                    class="small-button"
+                                    @click="pgChangePage"
+                                    :disabled="pgBtnIsDisabled('first')"
+                                >
+                                    &ltcc;&ltcc;
```
Nikomus: These won't make sense to a screen reader

charlesstrange2: added some aria bits;

alternatives - use a `<nav>` tag structure ?

PR #272, opened 2023-03-09
1.6/1935 add a spinner to the trader page for large amounts of data vouchers displayed vvv ee

lauragurney:
https://trello.com/c/SP7VGSAQ/1935-add-a-spinner-to-the-trader-page-for-large-amounts-of-data-vouchers-displayed-vvv-ee

### DIFF
--- a/src/pages/Account.vue
+++ b/src/pages/Account.vue
@@ -1,12 +1,5 @@
 <template>
     <div>
-        <div id="look here">
-            <spinner
-                v-bind:active="spinnerActive"
-                text="Please wait, working on it..."
-            />
-        </div>
-
         <main class="container fullwidth" id="account">
             <div class="content fullwidth">
                 <h1>Requested Payments</h1>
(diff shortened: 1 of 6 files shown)


src/pages/Spinner.vue
```diff
@@ -0,0 +1,74 @@
+<template id="spinner">
+    <div v-if="active" class="spinner-wrapper">
+        <div class="spinner">
+            <div></div>
```
charlesstrange2, line 4: The use of these divs to create height is a bit of a sledgehammer;
wd cna probably solve that with css.

src/pages/Spinner.vue
```diff
@@ -0,0 +1,74 @@
+<template id="spinner">
```
charlesstrange2, line 1: Spinner.vue probably belongs under /components rather than /pages

src/pages/Account.vue
```diff
@@ -10,190 +10,205 @@
                         v-bind:state="message.state"
                     ></message>
                 </transition>
+                <div v-if="spinnerActive">
+                    <spinner
+                        v-bind:active="spinnerActive"
+                        text="Please wait, working on it..."
+                    />
+                </div>
+                <div v-if="!spinnerActive">
```
charlesstrange2: Could also be "v-else", which when applied to the next element after a "v-if", will act as an else for. It.

src/pages/Account.vue
```diff
@@ -293,13 +308,21 @@ export default {
...
         },
         pgChangePage: function (event) {
+            this.loadSpinner();
             const key = event.target.id;
             const pg = this.voucherPagination || {};
             const page = pg.hasOwnProperty(key) ? pg[key].page : 1;
             Store.getVoucherPaymentState(page);
         },
+        loadSpinner: function () {
+            this.showSpinner();
```
charlesstrange2: Currnently this looks like
- mount the page
- show the spinner
- set a timeout to hide th spinner at 2 aeconds
- fetch the data <- may take more than 2 seconds
- time out triggers
- data may not have arrived.

It's OK for now, but we should remember to tie spinner hiding to data-has-been-fetched when we fix that bit.

And I don't have a concrete idea about that at the moment.
I'm thinking
- Store has a flag that says it's fetching for us
- we watch that in data here
- when it changes to true, we start a timer to show spinner
- when it goes false, we cancel timer and hide spinner. 
- if it takes longer than timer, spinner shows
But that's for another card.

charlesstrange2 (changes requested):
Try the v-else, cull some divs and move the vue file location

src/pages/Spinner.vue
lauragurney: Removing them means there is no space for the spinny circle so it just disappears entirely. I checked other sources and they all said 4 x divs to allow for the spinner.

src/pages/Spinner.vue
lauragurney: Good shout, done.

src/pages/Account.vue
lauragurney: Done

charlesstrange2 (approved):
♻️

PR #273, opened 2023-03-23
1.6/1938 message when scanning vouchers vvv eee queue

charlesstrange2:
So much stuff here :-(

- the IDE es6-ified a bunch of stuff :-(
- abstracted the submit vouchers button out into it's own component, and pulled it's SCSS into there from app.scss
- replaced that in Queue, Payment, Tap and Scan
- fixed a problem in those pages with the Mixins
- disabled SPA reloads when in dev mode to avoid a conflict with webpack.
- pulled fonts definition into it's own SCSS file to avoid a problem with referencing font locations in components.
- refined the generation of NetMgr into a factory function, so we can make another instance.
- used that instance to add queue capabilities if we get a 202.


src/components/Queue.vue
```diff
@@ -112,15 +98,14 @@ export default {
...
+        if (this.queue.sendingStatus) {
             this.startSpinner();
         }
     },
     computed: {
         currentlyShown: function () {
             return (
-                this.fail ||
-                (this.vouchers.length >= 1 && !Store.getVouchersOnlineStatus())
+                this.vouchers.length >= 1 && !Store.getVouchersOnlineStatus()
```
neonleann, line 108: Are you quite sure you don't need the brackets here?

src/pages/Tap.vue
```diff
@@ -273,10 +224,10 @@ export default {
             }
         },
         getKeyCharCode: function (event) {
-            // Try to cross platform catch the keycode
+            // Try to cross-platform catch the keycode
             // Note, there's also "event.which" (int)
             // There's also "event.key" (string), which MDN thinks is better;
-            const charCode = event.keyCode ? event.keyCode : event.charCode;
+            const charCode = event.keyCode ?? event.charCode;
```
neonleann, line 230: So we're getting null if there's no charCode?? Is that okay?

neonleann:
I'm not seeing any new tests for this (you know I love tests). Are there already some that cover it and still pass?

src/pages/Tap.vue
charlesstrange2: isn't it saying "use keycode, unless falsy, then use charcode"?

charlesstrange2:
yeah, the tests are just broken, sadly :-(

src/pages/Tap.vue
neonleann: You're probably right. I'm still not totally au fait with the old double question mark. As you were.

src/components/Queue.vue
charlesstrange2: I think so; they were protecting the || and && conditions from each other.
without the || I think it's ok

neonleann: I have done some research. From what I can tell, since the AND operator gets precedence over the OR operator, this will probably be fine. But I did enjoy this comment on the subject of losing brackets: "It matters less whether you are confident in your grasp of the language. What matters more is the grasp of the language of the n00b that follows you." But I shall approve and leave it up to you.

neonleann (approved):
Approved (but with a nagging worry about tests and brackets). I trust you Charlie!

PR #277, opened 2023-04-11
css tweaks to allow iPhone 8 to see queued vouchers

lauragurney:
https://trello.com/c/AcLijeoK/1999-iphone-unable-to-see-queued-vouchers-obscured-by-footer-on-simulator

This just allows iPhone 8 users to view the queued vouchers as before the list was hidden behind the footer. Here is what it would look like on an iPhone 8 now:
Screenshot 2023-04-11 at 15 59 11


charlesstrange2 (approved):
slim-line!

PR #289, opened 2023-06-28
fix: source map loader

tobybatch:
Gets rid of the annoying yellow warning source map error, and loads usable JS in the console. ESLint when a bit OTT, the actual non - linting changes are

Line 4
Line 150 -155
Line 190 - 192


karalan (approved):
🦑

